PR #52, opened 2023-03-13
Generate deterministic GUIDs for P&L rows

rich-exogee:


skeep-exogee:
Lets leave this PR open until we need to work on Xero again. We don't want to lose all the hard work that has gone into this PR.

PR #64, opened 2023-05-05
Getting Started Tidy Up

skeep-exogee:
This PR:

- [x] Make the "Create Graphweaver" cli tool work and start the server successfully
- [x] Consolidate the CLI tools into a single package
- [x] Remove dependency on mikro postgres when not using postgres!


src/packages/cli/src/create/index.ts
```diff
@@ -0,0 +1,68 @@
...
+					{
+						value: Backend.REST,
+						name: 'REST Backend',
+					},
+				],
+			},
+			{
+				type: 'confirm',
+				name: 'createDirectory',
+				message: `OK, we're ready- I'm going to create a new app in "${cwd()}" - is that OK?`,
```
kyelewis, line 51: Is this message correct now, i think it will create in cwd() + projectName?

kyelewis (approved):
generally looks good, just not sure the path we confirm is correct?

src/packages/cli/src/create/index.ts
skeep-exogee: Good spot added:

`OK, we're ready- I'm going to create a new app in "/Users/skeep/Projects/exogee/graphweaver/src/packages/cli/todo" - is that 
OK? (Y/n) `

PR #66, opened 2023-05-10
Feat/custom fields

skeep-exogee:
This PR adds support for custom fields in the table. This is shown in the search column below:

Screenshot 2023-05-10 at 3 05 04 pm

The example is off a link to google to search for the star wars character name


src/packages/vite-plugin-graphweaver/src/loaders/custom-fields.ts
```diff
@@ -0,0 +1,25 @@
+import path from 'path';
+import fs from 'fs/promises';
+
+export const loadCustomFields = async (configPath: string) => {
+	try {
+		let customFieldsPath = path.resolve(process.cwd(), 'src', 'admin-ui', 'custom-fields');
+
+		try {
+			const { adminUI } = await import(configPath);
+			if (adminUI?.customFieldsPath) customFieldsPath = adminUI.customFieldsPath;
```
kyelewis, line 10: Can these just be another export from the existing admin UI file? Or is there a technical reason that doesn't work

kyelewis (commented):
I really want this feature! So just a couple of ergnonomic questions, i'll have to have a proper read

kyelewis (approved):

PR #70, opened 2023-05-19
Update default ports and path, and tidy up CLI logging

kyelewis:
This PR:
- Makes the default frontend port  9000
- Makes the default backend port (frontend + 1)
- Makes the default route "/"

8000 and 3000  basically conflict with the defaults on everything, so less conflicts to resolve will make life easier.

---

If we're going to keep using serverless-offline, should we consider a small local patch to its startup logging, so we can remove some of the extra, confusing info it prints to the console at startup?

The trace logging is OK, but all the URLs and routes are a bit extra.


skeep-exogee (commented):
Can you check these places I think they will also need to be updated:

https://github.com/exogee-technology/graphweaver/blob/22d1c9280ab5f776290a2df0378dddddf78b4e1d/src/examples/xero/README.md#L33-L34

https://github.com/exogee-technology/graphweaver/blob/5a6746c68cdf9aa3b6b6e0a03113e5a4c99fb62f/src/packages/apollo/src/plugins/cors.ts#L41

https://github.com/exogee-technology/graphweaver/blob/7b968991334d7f047166663741b8f112a6b4e791/.vscode/launch.json#L40

https://github.com/exogee-technology/graphweaver/blob/5a6746c68cdf9aa3b6b6e0a03113e5a4c99fb62f/src/examples/xero/.env.example#L3

https://github.com/exogee-technology/graphweaver/blob/e611fdd33f05332394ba8c6d0a5cf41e41ed84fc/src/packages/cli/src/vite-config.ts#L19

kyelewis:
- [ ] `packages/apollo/src/plugins/cors.ts`

- [ ] `examples/xero/.env.example`

- [ ] `graphweaver/.vscode/launch.json`
 I can edit this one but i don't use vscode so i'd be flying blind on it
 
- [ ] `cli/src/vite-config.ts`

skeep-exogee (approved):

PR #73, opened 2023-05-24
Allow custom cors validation logic

kyelewis:


src/packages/apollo/src/plugins/cors.ts
```diff
@@ -1,7 +1,22 @@
 import { ApolloServerPlugin } from '@apollo/server';
 import { logger } from '@exogee/logger';
 
-export const Cors: ApolloServerPlugin = {
+// @todo does anyone need this legacy code?
+const defaultValidateOrigin = (origin: string) => {
+	const rootDomain = process.env.ROOT_DOMAIN;
```
skeep-exogee, line 6: I dont think we use this, shall we remove it?

skeep-exogee: im not sure what `easy-` is

kyelewis: happy to remove it, it's just the logic in the existing code

skeep-exogee: yeah i think lets tidy it up we can always add back if its needed

src/packages/apollo/src/plugins/cors.ts
```diff
@@ -1,7 +1,13 @@
 import { ApolloServerPlugin } from '@apollo/server';
 import { logger } from '@exogee/logger';
 
-export const Cors: ApolloServerPlugin = {
+export interface CorsPluginOptions {
+	validateOrigin?(origin: string): boolean;
+}
+
+export const corsPlugin = ({
+	validateOrigin = () => false,
```
kyelewis, line 9: @skeep-exogee do you think the default should be to allow or deny?

skeep-exogee: yeah as it forces us to supply it 👍

kyelewis: great, then validate = false is the default we want

skeep-exogee (approved):

PR #75, opened 2023-05-24
Feat/auth

skeep-exogee:
Recording 2023-05-24 at 14 14 52

What this PR has:

- Create a local provider concept where a username and password can be used to access the API
- Creates a login mutation that can be used to login and get a JWT
- A login form that is redirected to when a forbidden error occurs for a guest
- JWT with secret signing
- Restrict access to the admin ui metadata using a hook
- Guest access to the login mutation

What it does not have:

- Stateful JWT
- Ability to logout either in the UI or API
- JWT with cert signing
- A refresh token mechanism
- Storage of passwords as hashes
- Storage of credentials in the db
- Error display and validation on the login form
- A way to register a user
- Forgotten password flow / reset password flow
- Email verification
- MFA
- Cognito, Auth0 or Okta login
- Definition of a guest and authorized roles exported by the auth package
- Credential stuffing / Brute force protection


src/packages/auth/src/authentication/providers/local/plugin.ts
```diff
@@ -0,0 +1,97 @@
...
+			});
+			upsertAuthorizationContext(contextValue);
+		} else {
+			// Case 2. There is an auth header
+			logger.trace('Got a token, checking it is valid.');
+
+			const tokenProvider = new LocalAuthTokenProvider();
+
+			try {
+				const decoded = await tokenProvider.decodeToken(authHeader);
```
kyelewis, line 56: We should send the Authorization header as `[mode] [credential]` ie `Bearer (token)` to keep to spec -  should we split that here or in the token provider?

src/packages/auth/src/user-profile.ts
```diff
@@ -0,0 +1,44 @@
+import { BaseDataEntity } from '@exogee/graphweaver';
+import { AuthProvider } from './authentication/base-auth-token-provider';
+
+export type UserProfileData = {
+	provider: AuthProvider;
+	id?: string;
+	displayName?: string;
+	name?: {
```
kyelewis, line 8: is the name object breakdown necessary/useful to us?

kyelewis (commented):
Small comments/feedback

src/packages/auth/src/user-profile.ts
skeep-exogee: I just wanted to give maximum flexibility as you never know what the structure of a user might be.

src/packages/auth/src/authentication/providers/local/plugin.ts
skeep-exogee: yeah ill add this

skeep-exogee: all done

kyelewis (approved):

PR #94, opened 2023-06-20
Example/column level security

tnodell:
Implements an example of Column Level Security on the MySQL example workspace.


src/examples/rest/src/backend/database.ts
```diff
@@ -9,7 +9,7 @@ export const myConnection = {
 		driver: MySqlDriver,
 		dbName: 'todo_app',
 		user: 'root',
-		password: 'password',
+		password: '',
```
skeep-exogee, line 12: Whilst we are here can we replace these with the env vars?

in the .env example:

MYSQL_USERNAME="root"
MYSQL_PASSWORD="password"

src/examples/rest/src/backend/schema/task/entity.ts
```diff
@@ -34,6 +35,43 @@ const acl: AccessControlList<Task, AuthorizationContext> = {
...
+		},
+		LOW: {
+			description: 'LOW',
+		},
+	},
+});
+
+type TaskField = keyof InstanceType<typeof Task>;
+
+export const preventLightSideAccess = (
```
skeep-exogee, line 55: Maybe just add a longer comment of what is going on here about how Luke is part of the Light side and is unable to see the priority column etc, etc

skeep-exogee (approved):

PR #95, opened 2023-06-26
Auto Generated Types

skeep-exogee:
This feature allows an Developer to start a watch process within their GraphWeaver project that will watch for GraphQL schema queries and mutations such as:

```
gql`
	query Tenants {
		tenants {
			id
			tenantName
		}
	}
`;
```

When working on a project you can start the watch feature by running `pnpm watch`. When the watch process finds a query or mutation a `*.generated.tsx` file will appear next to the file with the GraphQL specified. For example:

component.tsx
component.generated.tsx 

This auto generated file has strongly typed Apollo Client use hooks in it. For example the above query would have the following hook auto generated:

```
import { useTenantsQuery } from './index.generated';

...

// To know nav links we need to know the tenants.
const { data } = await useTenantsQuery();

if (!Array.isArray(data.tenants)) return;

return [
	{ name: 'All Companies', route: '/xero-dashboard' },
	...data.tenants.map((tenant) => ({
		name: tenant.tenantName,
		route: `/xero-dashboard/${tenant.id}`,
	})),
];

```

In the above example `data` is now strongly typed.


src/packages/builder/src/start/frontend.ts
```diff
@@ -1,46 +1,56 @@
...
-		`🚀 Admin UI: ${
-			server.resolvedUrls?.local?.[0] || server.resolvedUrls?.network?.[0] || 'Could not get URL'
-		}`
-	);
+	// We can now generate the front end functions and types
+	await codeGenerator();
+
+	// Let's check if we need to start the server
+	if (server) {
+		server.restart();
```
skeep-exogee, line 20: There is an issue at the moment with vite not being able to watch for the changes to the generated files. Im trying to look into a solution that is not just restart vite!

rich-exogee (approved):

PR #96, opened 2023-06-27
Admin UI Metadata Resolver Error

skeep-exogee:


src/examples/rest/src/backend/schema/task/entity.ts
```diff
@@ -94,6 +94,9 @@ export class Task extends GraphQLEntity<OrmTask> {
 	@Field(() => Priority, { nullable: true })
 	priority?: Priority;
 
+	@Field(() => [String])
+	test = ['test'];
```
skeep-exogee, line 98: Im going to remove this but this is what I used to replicate the error. Does this look correct @kyelewis

src/packages/apollo/src/metadata-service/resolver.ts
```diff
@@ -74,10 +74,8 @@ export const getAdminUiMetadataResolver = (hooks?: AdminMetadata['hooks']) => {
 							type: entityName,
 						};
 						const relatedObject = objectTypeData[entityName];
-						if (field.typeOptions.array) {
-							if (!relatedObject) {
-								throw new Error(`Unknown entityName ${entityName}`);
```
skeep-exogee, line 79: I think it is reasonable not to throw here as it is legit to have primitives in an array.

kyelewis (approved):

PR #97, opened 2023-06-29
Datasource introspection

skeep-exogee:
This PR includes two major pieces of functionally:

- You can now run `npx graphweaver import postgres|mysql|sqlite` to import a datasource and all the files are created for you
- There is an end to end package that now:
   - Creates a new GW app using the CLI
   - imports an sqlite data source
   - runs a basic query and mutation against the app created

This all runs in the github action `end-to-end-tests`

The following is also included:

- [x] fix nullable on relationship field
- [x] import enum to graphql file
- [x] create schema index file for each entity
- [x] create resolver file for each entitiy
- [x] export all data entities from entities index
- [x] export all resolvers from schema index
- [x] create database.ts file and export connection
- [x] data entity add relationship and raw id attribute
- [x] fix json scalar type
- [x] fix manytomany decorator
- [x] `geography(Point,4326)` type in schema entity file for depot (workaround using String)
- [x] `@Field(() => String[]` ---> `@Field(() => [String]` in schema entity file
- [x] fix manyToOne relationships
- [x] create test example project
- [x] auto import mikroorm
- [x] auto import scalar if used
- [x] add message to CLI of what to do once the import is complete
- [x] create project from cached example
- [x] input db settings into CLI
- [x] Run against a MySQL database
- [x] Fix `Error: DateScalar can only serialize string values`
- [x] Fix `Error: JSONObject cannot represent non-object value: [1,2,3,4,5,6,7]`
- [x] Fix `Error: Objects are not valid as a React child (found: object with keys {type, driverId, businessUnitName})`
- [x] Fix `Error: Entity 'BusinessUnit' does not have property 'assignedTipSitesInverseInverse'`
- [x] Fix `Error: Enum "TipTicketManualGrossWeightReason" cannot represent value: ""`
- [x] Fix `Error: Objects are not valid as a React child (found: object with keys {keys, endpoint, expirationTime})` on Subscription table
- [x] Fix `Error: Response not successful: Received status code 400` on InternalMetadata
- [x] Feedback when importing from the CLI
- [x] Rename primary key to `id` if named differently
- [x] Invalid connection details
- [x] Empty database
- [x] Unsupported data source
- [x] Mask password entry
- [x] auto add summary field
- [x] Filters error `Cannot include a '?' character in a manually specified `to.pathname` field [{"pathname":"/Album/graphweaver-admin-new-entity?filters=eyJUaXRsZSI6eyJUaXRsZSI6IkxldCBUaGVyZSBCZSBSb2NrIn19"}]. Please separate it out to the `to.search` field. Alternatively you may provide the full path as a string in <Link to="..."> and the router will parse it for you.`
- [x] Unable to edit with `Variable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"Name\" of required type \"String!\" was not provided., \nVariable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"Milliseconds\" of required type \"Float!\" was not provided., \nVariable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"UnitPrice\" of required type \"String!\" was not provided.`
- [x] Unable to add new with `Variable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"Name\" of required type \"String!\" was not provided., \nVariable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"Milliseconds\" of required type \"Float!\" was not provided., \nVariable \"$_v0_input\" got invalid value {} at \"_v0_input.data[0]\"; Field \"UnitPrice\" of required type \"String!\" was not provided.`


src/examples/databases/src/backend/schema/user/entity.ts
```diff
@@ -1,5 +1,4 @@
-import { GraphQLEntity, SummaryField } from '@exogee/graphweaver';
-import { Field, ID, ObjectType } from 'type-graphql';
+import { GraphQLEntity, SummaryField, Field, ID, ObjectType } from '@exogee/graphweaver';
```
tnodell, line 1: All of these "Field, ID, ObjectType, Resolvers" etc are they exactly the same thing just coming through gw now?

skeep-exogee: yeah, I did this for two reasons:

1. one less thing to explain to a new user
2. If we wanted to overwrite one with our own custom version we could and there would be no client changes needed.

what do you think?

tnodell: Yup that makes sense to me!

src/packages/admin-ui/src/pages/list/component.tsx
```diff
@@ -97,10 +97,26 @@ export const List = () => {
...
+	// If we have an array as a value then we need to convert this to a string to display in the table
+	const rows = (data?.result ?? []).map((row) => {
+		// Hold any overrides we need to apply
+		const overrides: { [k in keyof typeof row]?: unknown } = {};
+		const { fields } = entityByName(entity);
+		for (const key in row) {
+			const field = fields.find((field) => field.name === key);
+			if (field?.type === 'JSON') {
+				// We have an array let's stringify it so it can be displayed in the table
+				overrides[key as keyof typeof row] = JSON.stringify(row[key as keyof typeof row]);
```
tnodell, line 109: JW Can we save these types as "values"?
as in `type OverrideKey = key as keyof typeof row`

src/packages/cli/package.json
```diff
@@ -21,12 +22,11 @@
...
 		"@exogee/graphweaver-admin-ui": "workspace:*",
 		"@exogee/graphweaver-builder": "workspace:*",
-		"@octokit/core": "4.2.1",
-		"@octokit/plugin-throttling": "6.0.0",
-		"@octokit/rest": "19.0.11",
+		"@exogee/graphweaver-mikroorm": "workspace:*",
 		"chokidar": "3.5.3",
 		"esbuild": "0.15.5",
 		"inquirer": "9.2.1",
+		"ora": "5.4.1",
```
tnodell, line 29: Nice! A terminal spinner.

skeep-exogee: Postgres DBs can take quite a bit of time im not sure why. SQLite and MySql are super quick

src/packages/admin-ui/src/pages/list/component.tsx
skeep-exogee: yeah good idea!

src/packages/cli/src/init/index.ts
```diff
@@ -32,13 +39,23 @@ export const initGraphWeaver = (projectName: string, backends: Backend[], versio
...
-export const init = async ({ version }: InitOptions) => {
+export const init = async ({ version, name, backend }: InitOptions) => {
 	console.log(`GraphWeaver ${version ? 'using version ' + version : ''}\n`);
 
-	import('inquirer').then(async ({ default: inquirer }) => {
-		const { projectName, createDirectory, backends } = await inquirer.prompt([
+	if (backend && name) {
+		initGraphWeaver(name, [backend], version);
+	} else {
+		const { default: inquirer } = await import('inquirer');
```
tnodell, line 52: You can await an import???

src/packages/cli/src/init/template.ts
```diff
@@ -74,8 +74,19 @@ export const makeDatabase = (projectName: string, backends: Backend[]) => {
 	},
 };`;
 
+	const liteDriverImport = `import { SqliteDriver } from '@mikro-orm/sqlite';`;
+	const liteConnection = `export const liteConnection = {
+	connectionManagerId: 'sqlite',
+	mikroOrmConfig: {
+		driver: SqliteDriver,
```
tnodell, line 81: Is this also the first time we're supporting SQLLite?

src/packages/core/src/base-entity.ts
```diff
@@ -25,7 +25,7 @@ export const AdminUISettingsMap = new Map<string, AdminUISettingsType>();
 
 @ObjectType()
 export class GraphQLEntity<D extends BaseDataEntity> {
-	public id?: string;
+	public id?: string | number;
```
tnodell, line 28: Why are we now supporting ids as numbers? Usually they're like "1"

src/packages/end-to-end/README.md
```diff
@@ -0,0 +1,20 @@
...
+- Imports a data source to be used to query
+- Makes requests to the API and tests the results
+
+## Database
+
+The database used to run these tests can be found here:
+
+https://github.com/lerocha/chinook-database
+
+The SQLite file has been copied from the above directory and stored in `database.sqlite`.
```
tnodell, line 15: Is it icky to store a db as a file in a project? I've never encountered keeping the db in the project before but it could be fine b/c its just an example.

src/packages/core/src/base-entity.ts
skeep-exogee: it depends on the data source they could be strings or numbers. We normally store as BigInt which has to be represented as a string in JS. But another DB might just use `int` which can be represented as a number.

src/packages/end-to-end/README.md
skeep-exogee: This is ok for this SQLite databases are files 👍

src/packages/end-to-end/package.json
```diff
@@ -0,0 +1,22 @@
+{
+	"name": "@exogee/graphweaver-end-to-end",
+	"version": "0.0.1",
+	"description": "Graphweaver Test Suite",
+	"license": "MIT",
```
tnodell, line 5: The license at the root is "license": "UNLICENSED",
What's the license of GW? and does it need to be consistent within packages?

src/packages/cli/src/init/template.ts
skeep-exogee: yeah!

src/packages/cli/src/init/index.ts
skeep-exogee: yes sir 😄

src/packages/end-to-end/src/__tests__/basic-mutation.test.ts
```diff
@@ -0,0 +1,27 @@
...
+						createAlbum(data: $data) {
+							id
+						}
+					}
+				`
+			)
+			.variables({ data: { ArtistId: { id: 1 }, Title: 'string' } })
+			.expectNoErrors();
+
+		expect(data?.createAlbum?.id).toBe('348');
```
tnodell, line 25: Is this id expected just because we know there are 347 other albums already present? Does this test fail if you run it twice?

src/packages/end-to-end/src/__tests__/basic-query.test.ts
```diff
@@ -0,0 +1,26 @@
...
+					query {
+						albums {
+							id
+						}
+					}
+				`
+			)
+			.expectNoErrors();
+
+		expect(data?.albums).toHaveLength(347);
```
tnodell, line 24: Similar to the above, looks like we're expecting this to just build and run once. If so, all good.

src/packages/end-to-end/package.json
skeep-exogee: everything should be MIT. I have updated the one in root.

src/packages/end-to-end/src/__tests__/basic-mutation.test.ts
skeep-exogee: each time the test is run the database is reset. So that the number is always 348.

src/packages/end-to-end/src/__tests__/basic-query.test.ts
skeep-exogee: yeah same as above 👍

src/packages/end-to-end/src/types.ts
```diff
@@ -0,0 +1,3 @@
+export type Album = {
+	id: number;
```
tnodell, line 2: I see this id is a number, why? SQlLite?

src/packages/end-to-end/tsconfig.json
```diff
@@ -0,0 +1,109 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
```
tnodell, line 5: Why is all this commented out?

skeep-exogee: thats the default config after `npx tsc --init`

src/packages/end-to-end/src/types.ts
skeep-exogee: yeah it is an integer in the sqlite db

src/packages/mikroorm/src/introspection/files/schema-entity-file.ts
```diff
@@ -0,0 +1,262 @@
...
+			if (prop.enum) {
+				const enumClassName = this.namingStrategy.getClassName(
+					this.meta.collection + '_' + prop.fieldNames[0],
+					'_'
+				);
+				enumDefinitions.push(this.getEnumClassDefinition(enumClassName));
+			}
+		});
+
+		let ret = '';
```
tnodell, line 66: What does ret mean?

src/packages/mikroorm/src/introspection/generate.ts
```diff
@@ -0,0 +1,234 @@
...
+			if ([ReferenceType.MANY_TO_ONE, ReferenceType.ONE_TO_ONE].includes(prop.reference)) {
+				prop.wrappedReference = true;
+			}
+		}
+	}
+};
+
+const generateSingularTypeReferences = (metadata: EntityMetadata[]): void => {
+	for (const meta of metadata.filter((m) => !m.pivotTable)) {
+		meta.className = pluralize.singular(meta.className);
```
tnodell, line 114: If this is "eggs" then it becomes "egg", but if it's "egg" it stays "egg"? Is this because we don't know if the className is plural in the db?

src/packages/mikroorm/src/introspection/generate.ts
```diff
@@ -0,0 +1,234 @@
...
+		for (const prop of meta.relations) {
+			prop.type = pluralize.singular(prop.type);
+		}
+	}
+};
+
+const convertSchemaToMetadata = async (
```
tnodell, line 121: We create the schema first and then populate the metadata?

src/packages/mikroorm/src/introspection/generate.ts
```diff
@@ -0,0 +1,234 @@
...
+
+		const config = database.em.config;
+		const driver = database.em.getDriver();
+		const platform = driver.getPlatform();
+		const namingStrategy = config.getNamingStrategy();
+		const connection = driver.getConnection();
+
+		console.log('Fetching database schema...');
+		const schema = await DatabaseSchema.create(connection, platform, config);
+		console.log('Building metadata...');
```
tnodell, line 194: Yes ^^

tnodell (approved):
Lots of tiny things but overall looks good! Woo introspection!

src/packages/mikroorm/src/introspection/generate.ts
skeep-exogee: correct it might be "eggs" as the tablename

src/packages/mikroorm/src/introspection/generate.ts
skeep-exogee: yeah we turn the database schema into metadata.

src/packages/mikroorm/src/introspection/files/schema-entity-file.ts
skeep-exogee: supposed to be the `ret`urned string but I dont like it now 😄 have renamed them all to `file`

PR #98, opened 2023-06-30
Feature/graphql dedupe

tnodell:
https://app.clickup.com/t/2704425/ID-4246

### DIFF
--- a/src/packages/apollo/src/plugins/graphql-deduplicator.ts
+++ b/src/packages/apollo/src/plugins/graphql-deduplicator.ts
@@ -5,12 +5,13 @@ export const dedupeGraphQL: ApolloServerPlugin = {
 	async requestDidStart() {
 		return {
 			async willSendResponse(requestContext) {
-				if (requestContext.response.body.kind == 'single') {
-					if (requestContext.response.body.singleResult.data) {
-						// TODO: check if enabled
+				if (
+					requestContext.response.body.kind == 'single' &&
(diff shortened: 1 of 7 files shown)


src/packages/apollo/src/plugins/graphql-deduplicator.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const dedupeGraphQL: ApolloServerPlugin = {
+	async requestDidStart() {
+		return {
+			async willSendResponse(requestContext) {
+				if (
+					requestContext.response.body.kind == 'single' &&
+					requestContext.response.body.singleResult.data
+				) {
+					if (typeof requestContext.response.body.singleResult.data.result === 'object') {
```
skeep-exogee, line 12: is this to check if its an Array?

if so, you could use:

`if (Array.isArray(requestContext.response.body.singleResult.data.result)) {`

src/packages/apollo/src/plugins/graphql-deduplicator.ts
```diff
@@ -0,0 +1,21 @@
...
+						requestContext.response.body.singleResult.data.result = deflate(
+							requestContext.response.body.singleResult.data.result as object
```
skeep-exogee, line 14: also this type might change based on the conversation above.

skeep-exogee (approved):

PR #103, opened 2023-07-14
Fix/filter-by-child

tnodell:
https://app.clickup.com/t/2704425/ID-4539


tnodell:
Closing b/c we suspect that the entire function containing `query.leftJoin(${table}.${key}, joinKey);`. We are close to deploying a test suite. When that is live we will remove the function and run the tests. If all the tests are successful then we know we don't need that function.

PR #102, opened 2023-07-13
Feature/related entities filter

tnodell:
https://app.clickup.com/t/2704425/ID-4245

This implements filters on nested entities that rely on the mikroORM provider.

```
query TasksFilteringTags($filter: TagsListFilter) {
  tasks {
    id
    tags(filter: $filter) {
      name
      id
    }
  }
}
```
```
{
  "filter": {
    "name": "Advanced"
  }
}
```

PR #104, opened 2023-07-17
Feature/provider support declarations

tnodell:
Providers must now register what features they support:
filtering by root, parentByChild, childByChild, pagination, sorting.
Defaults are provided for exisiting providers.

Built off of: https://github.com/exogee-technology/graphweaver/pull/102

https://app.clickup.com/t/2704425/ID-4245


src/packages/mikroorm/src/base-resolver/provider.ts
```diff
@@ -95,6 +96,26 @@ export class MikroBackendProvider<D extends BaseDataEntity, G extends GraphQLEnt
 
 	public readonly supportsInFilter = true;
 
+	// Default backend provider config
+	public readonly backendProviderConfig: BackendProviderConfig = {
+		filter: {
+			root: true,
+			parentByChild: true,
+			childByChild: true,
```
skeep-exogee, line 104: so `childByChild` is to enable filtering on a child entity within the graphql query?

tnodell: Yea that's what is now supported by https://github.com/exogee-technology/graphweaver/pull/102 related entities filter. I'm open to better naming conventions.

src/packages/apollo/src/typegraphql-params.ts
```diff
@@ -0,0 +1,40 @@
...
+import { EntityMetadataMap } from '@exogee/graphweaver';
+import { getMetadataStorage } from 'type-graphql';
+import { ArgParamMetadata } from 'type-graphql/dist/metadata/definitions';
+
+// Type guard for params that are arg kind
+export const isArg = (param: any): param is ArgParamMetadata => {
+	return param.kind === 'arg';
+};
+
+export const removeInvalidFilterArg = () => {
```
skeep-exogee, line 12: We could add something like this above the function?

```
// This function checks to see if the data provider supports nested child filters
// If it does not support filtering then the filter is removed
```

skeep-exogee (approved):

PR #106, opened 2023-07-20
Fix/nested create

skeep-exogee:
This PR addresses the nested entities bug. To confirm that it is working three tests were added:

1. Create an entity and a single nested entity
2. Create an entity and an array of nested entities
3. Update an entity and create a nested entity

To complete this work some changes were needed to the datasource introspection package and these are also included in this PR.

https://app.clickup.com/t/2704425/ID-4383


src/examples/databases/src/backend/index.ts
```diff
@@ -3,11 +3,10 @@ import Graphweaver from '@exogee/graphweaver-apollo';
 import { connectToDatabase } from '@exogee/graphweaver-mikroorm';
 
 import { resolvers } from './schema';
-
 import { connections } from './database';
 
 const graphweaver = new Graphweaver({
-	resolvers: resolvers,
```
tnodell, line 10: 👍

src/packages/core/src/utils/create-or-update-entities.ts
```diff
@@ -125,32 +127,69 @@ export const createOrUpdateEntities = async <G extends WithId, D extends BaseDat
...
-					...node,
-					[key]: result,
-				};
+				if (isLinking(childNode)) {
+					// If it's a linking entity or an array of linking entities, nothing to do here
+				} else if (Array.isArray(childNode)) {
+					// If we have an array, we may need to create the parent first as children need reference to the parent
+
+					// As are updating the parent from the child, we can remove this key
+					delete node[key as keyof Partial<G>];
```
tnodell, line 136: I don't totally understand this. Are we deleting this item so that it doesnt appear in the next iteration of the loop?

src/packages/core/src/utils/create-or-update-entities.ts
```diff
@@ -125,32 +127,69 @@ export const createOrUpdateEntities = async <G extends WithId, D extends BaseDat
...
+						parent = fromBackendEntity(parentDataEntity, gqlEntityType);
+						parentId = parent?.id;
+					}
+					if (!parentId) {
+						throw new Error(`Implementation Error: No parent id found for ${relatedEntity.name}`);
+					}
+
+					// Add parent ID to children and perform the mutation
+					const childMeta = getMeta(relatedEntity.name);
+					const parentField = childMeta.fields.find((field) => field?.getType() === gqlEntityType);
```
tnodell, line 152: Maybe for later, but can we rename getType() to getGqlEntityType?

src/packages/end-to-end/src/__tests__/api/mutation/basic-create.test.ts
```diff
@@ -19,7 +19,7 @@ describe('basic mutation', () => {
 					}
 				`
 			)
-			.variables({ data: { ArtistId: { id: 1 }, Title: 'string' } })
```
tnodell, line 22: Looks like we followed through on changing the db naming convention as discussed in that meeting with Kevin.

tnodell (approved):
Nice fix!

src/packages/core/src/utils/create-or-update-entities.ts
skeep-exogee: yeah we should review that when we review the whole metadata piece 👍

src/packages/end-to-end/src/__tests__/api/mutation/basic-create.test.ts
skeep-exogee: yeah correct

src/packages/core/src/utils/create-or-update-entities.ts
skeep-exogee: The connection is in reverse when we have an array of children. So...

Artist has albums.

an Album has an artistId in the database. 

Therefore, we don't need to send the `albums[]` array when we create the Artist.

PR #107, opened 2023-07-20
Fix: Double Login

skeep-exogee:
Fixes the issue where you have to login twice on the admin ui Rest example

### DIFF
--- a/src/packages/admin-ui-components/src/apollo.ts
+++ b/src/packages/admin-ui-components/src/apollo.ts
@@ -3,19 +3,36 @@ import { inflate } from 'graphql-deduplicator';
 
 import { uri } from './config';
 
+const AUTH_TOKEN_LOCAL_STORAGE_KEY = 'graphweaver-auth';
+
 const httpLink = new HttpLink({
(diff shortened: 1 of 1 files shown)


src/packages/admin-ui-components/src/apollo.ts
```diff
@@ -3,20 +3,37 @@ import { inflate } from 'graphql-deduplicator';
...
 const httpLink = new HttpLink({
 	uri,
 });
 
+const isExpired = (token: string) => {
+	try {
+		const decodedJwt = JSON.parse(atob(token.split('.')[1]));
+		return decodedJwt.exp * 1000 < Date.now();
+	} catch (e) {
+		return true;
```
tnodell, line 17: Maybe log or throw an error here?

tnodell (approved):

src/packages/admin-ui-components/src/apollo.ts
skeep-exogee: good idea added a warning!

PR #115, opened 2023-07-24
Fix/sprint polish

skeep-exogee:
Polish from last sprint.

1. The filters scroll in the Admin UI but do not show the drop downs. The "solution" for now is to reduce the number of filters to 5 and remove the scrolling. This is a stopgap until we redesign the filters layout.
2. The pnpm lock file was still getting updated from the end to end test app (as the file copy process could be stopped during install). This PR now ignores that dir in the workspace and additional flags are added in the end-to-end package to both ignore the workspace and to freeze the lock file. 
3. The plugins for the GraphQL codegen were not imported correctly when the code was bundled. This is now fixed with a custom import function to import the correct plugin.


src/packages/admin-ui-components/src/filter-bar/component.tsx
```diff
@@ -51,7 +51,11 @@ export const FilterBar = ({ iconBefore }: { iconBefore?: ReactNode }) => {
 	const getFilterComponents = (entityName: string) => {
 		const rowEntity = entityByName(entityName);
 
-		return rowEntity.fields.map((field) => {
+		// @todo - currently the filters are not fitting on the screen
+		// we plan to redo this filter bar so that it is a drop down
+		// for now the workaround is to reduce the number of filters to 5
```
tnodell, line 56: Just need to communicate this to Neil/whoever's demoing so that we don't look afool should someone want >5 filters.

tnodell (approved):
👍

src/packages/admin-ui-components/src/filter-bar/component.tsx
skeep-exogee: Yeah, ran this past @thekevinbrown so he is aware.

PR #116, opened 2023-07-25
Feature/apollo connection register

tnodell:
Moves the connectionToDatabase into the backend as a part of minimal file size.

https://app.clickup.com/t/2704425/ID-4247


src/packages/core/src/base-resolver.ts
```diff
@@ -91,6 +91,13 @@ export function createBaseResolver<G extends WithId, D extends BaseDataEntity>(
 	const entityFields = metadata.fields.filter((field) => field.target === gqlEntityType);
 	const enumSet = new Set(metadata.enums.map((enumMetadata) => enumMetadata.enumObj));
 
+	const entityMetadata = {
```
skeep-exogee, line 94: is this being used? Could you use it inside both the `EntityMetadataMap` and for the `static metadata`?

src/packages/server/src/index.ts
```diff
@@ -94,6 +98,28 @@ export default class Graphweaver<TContext extends BaseContext> {
 			}
 		}
 
+		const resolvers = (this.config.resolvers || []) as any; // BaseResolverInterface[]
```
skeep-exogee, line 101: can we type this?

tnodell: Schema expects this Function[] thing

```
Type 'BaseResolverInterface<D, G>[]' is not assignable to type 'NonEmptyArray<Function>'.
  Type 'BaseResolverInterface<D, G>[]' is not assignable to type '[Function, ...Function[]]'.
```

src/examples/rest/src/backend/index.ts
```diff
@@ -16,7 +14,7 @@ export enum Roles {
 const graphweaver = new Graphweaver<AuthorizationContext>({
 	resolvers,
 	apolloServerOptions: {
-		plugins: [localAuthApolloPlugin(addUserToContext), connectToDatabase(myConnection)],
```
skeep-exogee, line 19: we also need to remove this plugin from the `cli init` template and the introspection template.

src/packages/core/src/base-resolver.ts
skeep-exogee: This could be `const entityMetadata: BaseResolverMetadataEntry<D>` then you can remove the cast on line 100?

src/packages/core/src/base-resolver.ts
```diff
@@ -306,6 +307,13 @@ export function createBaseResolver<G extends WithId, D extends BaseDataEntity>(
 
 	@Resolver()
 	abstract class BaseResolver implements BaseResolverInterface {
+		static metadata: {
```
skeep-exogee, line 310: this can be removed now?

tnodell: Yes, because the connection is in the constructor of the MikroBackendProvider class now

skeep-exogee (approved):

PR #125, opened 2023-08-02
Added alert box

siowhoi:
https://app.clickup.com/t/2704425/ID-4720

Screenshot 2023-08-02 at 11 54 34 am
Screenshot 2023-08-02 at 12 07 37 pm


src/packages/admin-ui-components/src/alert/alert.tsx
```diff
@@ -0,0 +1,34 @@
...
+		['warning']: <WarningIcon />,
+		['info']: <InfoIcon />,
+		['success']: <SuccessIcon />,
+	};
+
+	return (
+		<div className={classNames(styles.container, severityClass[severity])}>
+			{severityIcon[severity]}
+			<Spacer width={10} />
+			{children}
```
tnodell, line 31: We're passing just the text "Invalid username or password." it might be safer to pass this inside a span. @skeep-exogee ?

skeep-exogee: yeah wrapping in a span or div would be a good idea 👍

src/examples/rest/README.md
```diff
@@ -17,7 +17,7 @@ CREATE TABLE task (
   completed BOOLEAN NOT NULL DEFAULT false,
   created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-  user_id INT NOT NULL
+  user_id INT NOT NULL,
```
skeep-exogee, line 20: great catch!

src/packages/admin-ui-components/src/alert/styles.module.css
```diff
@@ -0,0 +1,45 @@
...
+    background-color: #abe9b8;
+    color: green;
+}
+
+.success svg {
+    fill: green;
+}
+
+.info {
+    background-color: #abbce9;
```
skeep-exogee, line 30: I wonder if we should start defining these colours as vars. What do you think?

src/packages/admin-ui-components/src/login/component.tsx
```diff
@@ -74,11 +80,7 @@ export const Login = ({ onLogin }: LoginProps) => {
 							Login
 						</Button>
 					</div>
-					{!!error && (
-						<div>
-							<b>Error: ${error.message}</b>
-						</div>
-					)}
+					{!!error && <Alert>Invalid username or password.</Alert>}
```
kyelewis, line 83: Any reason we're not providing the error message? Because we allow pluggable auth, there's no guarantee that 'invalid username or password' is the error.

src/packages/admin-ui-components/src/spacer/component.tsx
```diff
@@ -0,0 +1,40 @@
+import classNames from 'classnames';
+
+type Props = {
```
kyelewis, line 3: the typing here (`type Props`) is different from the typing in Alert (`interface AlertProps`), we should aim to keep it consistent.

kyelewis (commented):
I think we could look at the colour contrast and increase it- especially for the red exclaimation and the orange box and text are probably too similar foreground to background.

src/packages/admin-ui-components/src/alert/alert.tsx
siowhoi: fixed.

src/packages/admin-ui-components/src/alert/styles.module.css
siowhoi: Done.

src/packages/admin-ui-components/src/spacer/component.tsx
siowhoi: Fixed

src/packages/admin-ui-components/src/login/component.tsx
siowhoi: One of the acceptance criteria is to not display if the password or username is at fault. Should I put the error.message in the console.log?

skeep-exogee:
> I think we could look at the colour contrast and increase it- especially for the red exclaimation and the orange box and text are probably too similar foreground to background.

Yeah agree we are getting Kevin A to update the design  👍

skeep-exogee (approved):

src/packages/admin-ui-components/src/login/component.tsx
kyelewis: my read of that requirement is that the backend shouldn’t tell you whether the username or the password individually were the problem, i have a feeling that’s already the case but we could check.

siowhoi: You are right. The backend is already handling this. Fixed. I've updated it to display error.message.

PR #127, opened 2023-08-02
Create SECURITY.md

skeep-exogee:


SECURITY.md
```diff
@@ -0,0 +1,15 @@
+# Security Policy
+
+## Supported Versions
+
+Currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 0.x.x   | :white_check_mark: |
```
tnodell, line 9: what does this mean? Version 0.2.2 of graphweaver is supported by graphweaver?

tnodell (approved):

SECURITY.md
skeep-exogee: Yeah its supposed to mean that we are supporting all 0.x.x versions of the code with security updates.

For example, when we have version 1.x.x we might choose to no longer support security updates for 0.x.x.

PR #147, opened 2023-08-10
Add an example of a field resolver

skeep-exogee:


src/examples/rest/src/backend/schema/task/entity.ts
```diff
@@ -97,6 +98,11 @@ export class Task extends GraphQLEntity<OrmTask> {
 	@Field(() => Priority, { nullable: true })
 	priority?: Priority;
 
+	@Field((type) => String, { nullable: true })
+	slug(@Root() task: Task) {
+		return `${task.id}:${task.description}`;
```
tnodell, line 103: What is a "Field Resolver"? Aren't all of these field resolvers?

tnodell (approved):

src/examples/rest/src/backend/schema/task/entity.ts
skeep-exogee: I have tried to define it here https://graphweaver.com/docs/graphql-entities-and-resolvers#25e235177d7f4a698136d9f9fe2c349c

Does the doc make it clearer?

tnodell: Hmm I understand it now, just the names "Field Resolver" and "Field Decorator" look very similar.

```
@Field(() => String)
	description!: string;

@Field(type => String, { nullable: true })
slug(@Root() task: Task) {
	return `${task.id}:${task.description}`;
}

```

Issue #153, opened 2023-08-22
`graphweaver init` UX feedback


kyelewis:
Running `graphweaver init`:

- `Initing project undefined, undefined, undefined, undefined, undefined, undefined...` - Lots of undefined

- Command describes a `password`, `host`, `port` and `user` argument but does not use them.

- The `name` argument is ignored if used without `backend` also.

- If the `backend` argument is used with a valid backend but no `name` argument is provided, the prompts run doubled/twice.

- The `backend` argument cannot specify multiple backends, but the interactive flow allows this.

- Missing `database name` argument?

- The `What would your like to call your new project?` prompt is not validated for a valid `package.json` name, and the init fails later if left empty.

- `Make sure you to pnpm install, then pnpm start.` - The word `to` should change to `run`?

- `Make sure you to pnpm install, then pnpm start.` - Missing instruction to `cd` in to the project directory first.

- When choosing sqlite, `src/database.ts` has a lot of extra whitespace in it.

- When choosing sqlite, no clear immediate documentation on '%%REPLACE_WITH_DB_NAME%%' or links to other documentation.

kyelewis:
Working on this in https://github.com/exogee-technology/graphweaver/pull/156

siowhoi:
Closing this issue because all the issues have been addressed.

PR #156, opened 2023-08-22
`graphweaver init` UX improvements

kyelewis:
https://github.com/exogee-technology/graphweaver/issues/153 

- Update printing of CLI arguments for `init` and `import`
- Remove currently unused `init` arguments
- Default name and backends in the interactive flow if one is provided on CLI
- Allow specifying multiple backends on CLI
- Validate project name
- Update instructions to include to change to the project directory.
- Return error code 1 if there are missing packages on `import`


skeep-exogee (approved):
Looks good thanks @kyelewis !

PR #171, opened 2023-09-06
Admin Panel UI Tests

skeep-exogee:
This PR uses playwright to test the admin panel.

There is a single test that will run against the SQLite db to ensure that the webpage is functioning and that you can navigate to the Album table.


tnodell (approved):
LGTM!

PR #172, opened 2023-09-08
Feat/auth

steve-keep:
This PR will add a password challenge and the ground work to support other MFA challenges.

https://github.com/exogee-technology/graphweaver/assets/85143753/52a3e7b2-8aae-4090-b81b-d7684ec04c32


tnodell (approved):

src/packages/auth/src/authentication/providers/password/provider.ts
```diff
@@ -0,0 +1,68 @@
...
+		return payload;
+	}
+
+	async stepUpToken(existingTokenPayload: JwtPayload) {
+		if (!secret) throw new Error('PASSWORD_AUTH_JWT_SECRET is required in environment');
+		const expires = Math.floor((Date.now() + ms(mfaExpiresIn)) / 1000);
+
+		const token = jwt.sign(
+			{
+				...existingTokenPayload,
```
thekevinbrown, line 56: Should this also merge the `amr` values to ensure `AuthenticationMethod.PASSWORD` is in there? This would be needed if the primary auth was magic link, then password was the step up.

thekevinbrown (approved):
A couple tiny things to potentially look at if it suits you, nothing worth blocking over.

PR #192, opened 2023-11-16
UI Fix - Refresh the table after update/create and Display error

siowhoi:
1. Added toast to display error message
2. Refresh the table after update/create

### DIFF
--- a/src/examples/rest/.gitignore
+++ b/src/examples/rest/.gitignore
@@ -1,3 +1,4 @@
 # Working build files
 .esbuild
 .graphweaver
+*.pem
--- a/src/packages/admin-ui-components/src/toast/styles.module.css
+++ b/src/packages/admin-ui-components/src/toast/styles.module.css
@@ -1,0 +1,43 @@
+.toastContainer {
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	justify-content: center;
+}
+
+.divider {
+	border-left: 1px solid white;
+	height: 80%;
+	padding-right: 16px;
+}
+
+.iconContainer {
+	display: flex;
+	min-width: 32px;
+	cursor: pointer;
+	min-height: 20px;
+	max-height: 20px;
+}
+
+.iconContainer:hover {
+	opacity: 0.7;
+}
+
+.closeIconWrapper {
+	width: 100%;
+}
+
+.closeIconLeft {
+	height: 100%;
+	width: 2px;
+	margin-left: 12px;
+	background-color: #fff;
+	transform: rotate(45deg);
+}
+
+.closeIconRight {
+	height: 100%;
+	width: 2px;
+	background-color: #fff;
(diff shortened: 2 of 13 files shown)


src/examples/rest/.gitignore
```diff
@@ -1,3 +1,4 @@
 # Working build files
 .esbuild
 .graphweaver
+*.pem
```
tnodell, line 4: What is this?

src/packages/admin-ui-components/src/toast/styles.module.css
```diff
@@ -0,0 +1,43 @@
...
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+	justify-content: center;
+}
+
+.divider {
+	border-left: 1px solid white;
+	height: 80%;
+    padding-right: 16px;
```
tnodell, line 11: This spacing looks weird

src/packages/admin-ui/src/pages/list/graphql.ts
```diff
@@ -8,7 +8,7 @@ export const queryForEntityPage = (entityName: string, entityByType: (type: stri
 	const queryName = pluralName[0].toLowerCase() + pluralName.slice(1);
 
 	return gql`
-		query AdminUIListPage($filter: ${pluralName}ListFilter, $pagination: ${pluralName}PaginationInput) {
+		query AdminUI${entityName}ListPage($filter: ${pluralName}ListFilter, $pagination: ${pluralName}PaginationInput) {
```
tnodell, line 11: @steve-keep you should look at this pattern before we introduce it.

When creating a new entity we want to update the table so that the new entity is shown. We've opted to refetched the query that populates the table. This query was generically named, so it was refetching all the entities in all the tables. Now we've made it specific to the entity so we only fetch the table that's changed.

@siowhoi  did I summarize this correctly?

tnodell (commented):
@siowhoi I want @steve-keep to look at the pattern of naming the AdminUI query by entity before approval. Otherwise, all looks good.

src/examples/rest/.gitignore
siowhoi: Authorisation has been added to the REST example. If you follow the steps in the README.txt, it will generate .pem files.

src/packages/admin-ui-components/src/toast/styles.module.css
siowhoi: fixed

src/packages/admin-ui/src/pages/list/graphql.ts
siowhoi: Yes, that's correct.

dylanlt (approved):
Approving because this has been tested and Steve can weigh in later on the approach. This only affects AIA because it is the only consumer as far as I am aware.

PR #196, opened 2023-11-29
Added mailchimp example

siowhoi:
Added mailchimp example and unit test
Screenshot 2023-11-29 at 3 27 40 pm


src/examples/mailchimp/.env.example
```diff
@@ -0,0 +1,4 @@
+MAILCHIMP_API_KEY=
+MAILCHIMP_SERVER_PREFIX=
+MAILCHIMP_LIST_ID=7b86936890
```
tnodell, line 3: Should these ids be public?

src/examples/mailchimp/package.json
```diff
@@ -0,0 +1,61 @@
+{
+	"name": "@exogee/graphweaver-example-mailchimp",
+	"version": "0.2.6",
```
tnodell, line 3: I believe gw is on 0.2.7 now.

src/examples/mailchimp/src/backend/schema/category/data-entity.ts
```diff
@@ -0,0 +1,14 @@
+import { BaseDataEntity } from '@exogee/graphweaver';
+
+export class CategoryDataEntity implements BaseDataEntity {
```
tnodell, line 3: I think of entities in `/schema` as being graphql entities, and these look like base data entities. Is this how AIA does it?

src/examples/mailchimp/src/backend/schema/category/entity.ts
```diff
@@ -0,0 +1,13 @@
+import { Field, GraphQLEntity, ID, ObjectType, SummaryField } from '@exogee/graphweaver';
+
+import { CategoryDataEntity } from './data-entity';
+
+@ObjectType()
```
tnodell, line 5: Maybe just split these into 

/entities/category.ts
and
/schema/category.ts

?

This is a stylistic difference so it's up to you. I think we should be consistent with the other examples and keep data entities and schema entities in their own folder.

src/examples/mailchimp/src/backend/schema/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import { InterestResolver, createInterestResolver } from './interest';
+import { MemberResolver, createMemberResolver } from './member';
+
+export type ClientOptions = {
+	apiKey: string;
+	server: string;
+};
+
+export const mailingList: MailingListResolver =
+	process.env.MAILCHIMP_API_KEY &&
```
tnodell, line 12: Would be good to try to catch any of these missing env vars with a specific error.

src/examples/mailchimp/src/backend/schema/member/provider.ts
```diff
@@ -0,0 +1,81 @@
+import Mailchimp from '@mailchimp/mailchimp_marketing/src/ApiClient';
+import {
+	createProvider,
+	ItemWithId,
+	createPaginationOptions,
+	inMemoryFilterFor,
+} from '@exogee/graphweaver-helpers';
+import { ClientOptions } from '../index';
+import { MemberDataEntity, MemberStatus } from './data-entity';
+import md5 from 'md5';
```
tnodell, line 10: I think we use argon2 for encryption. Can we double check that this is this lib to use with @steve-keep ?

src/packages/graphweaver-helpers/src/provider/index.ts
```diff
@@ -82,7 +82,6 @@ export const createProvider = <Entity extends WithId, Context, DataEntity extend
 		}
 
 		_mapDataEntity(dataEntity: DataEntity): DataEntity {
-			console.log('_mapDataEntity ', this?.dataEntity);
```
tnodell, line 85: ty I think I left this here.

tnodell (approved):
Couple of small things to check before merging. Otherwise looks good!

src/examples/mailchimp/package.json
siowhoi: fixed

src/packages/graphweaver-helpers/src/provider/index.ts
siowhoi: This line has been removed after merging with main.

src/examples/mailchimp/src/backend/schema/category/data-entity.ts
siowhoi: Fixed. They are restructured to entities/category.ts and schema/category

src/examples/mailchimp/src/backend/schema/category/entity.ts
siowhoi: Fixed as suggested.

src/examples/mailchimp/src/backend/schema/index.ts
siowhoi: added check for missing environment variables

src/examples/mailchimp/src/backend/schema/member/provider.ts
siowhoi: Is argon2 the same as md5? Mailchimp require md5 hashing on the email field.

PR #199, opened 2023-12-04
@ReadOnly decorator on Entity will disable Create/Save buttons

siowhoi:
Screenshot 2023-12-04 at 3 57 05 pm
Screenshot 2023-12-04 at 3 57 11 pm


steve-keep (approved):
perfect 😄

PR #201, opened 2023-12-05
Field will not be displayed on Detail Panel if the field is marked with @ReadOnlyProperty() decorator

siowhoi:
1. Add decorator @ReadOnlyProperty() to the field. Expect the field not be shown on Detail Panel.

2. Decorator ExcludeFromInputTypes has been deleted from the project.


tnodell (approved):
Looks good to me!

PR #202, opened 2023-12-06
Feat/gitleaks

tnodell:
Creates a Pre Commit step to confirm that no secrets are being committed to code.

After merging, contributors will need to:

1. `brew install gitleaks`

2. Install pre-commit from https://pre-commit.com/#install

3. Auto-update the config to the latest repos' versions by executing `pre-commit autoupdate`

4. Install with `pre-commit install`

https://github.com/gitleaks/gitleaks


steve-keep (approved):
All looks good and I even tested upgrading to the latest 👍

PR #204, opened 2023-12-06
Added property to hide entity in decorator AdminUISettings

siowhoi:
To hide an entity from AdminUI, add this to the entity:
```
@AdminUISettings({
	entity: {
		hide: true,
	},
})
```


steve-keep (approved):
Looks Great!

PR #213, opened 2023-12-12
Added hideFromDisplay to adminUISettings decorator

siowhoi:
To hide a column from the Admin UI, add this decorator to the field:
`@AdminUISettings({
	hideFromDisplay: true,
})`


steve-keep (approved):
Awesome 😄

PR #217, opened 2023-12-12
Feature / Added ability for custom fields to show on form.

thekevinbrown:
We had custom fields already in the table, but not in the detail form.

This PR extends this ability so that custom fields can show in both the table, and the form.

**Breaking Change**: The arguments for custom fields have changed. Instead of just being `entity`, now it's `{ entity, context }` where context tells you the context you're rendering in.

If users specify `showOn` then the field only shows in the places they specify, while if they do not, it shows everywhere.

Users can also override default fields with custom fields by giving them the same name. At that point the custom field wins.

I also refactored the field components into separate files while I was in there, but this was not strictly related to this change.

### DIFF
--- a/src/examples/rest/src/admin-ui/custom-fields/link.tsx
+++ b/src/examples/rest/src/admin-ui/custom-fields/link.tsx
@@ -1,13 +1,20 @@
 import { MouseEventHandler } from 'react';
 
-import { Task } from '../../types.generated';
 import { ReactComponent as OpenIcon } from '../assets/16-open-external.svg';
+import { CustomFieldArgs } from '@exogee/graphweaver-admin-ui-components';
 
-export const Link = (task: Task) => {
+interface Task {
+	user: {
+		label: string;
+		value: string;
+	};
+}
+
+export const Link = ({ entity }: CustomFieldArgs<Task>) => {
 	const handleClick = (e: MouseEventHandler<HTMLDivElement>) => {
 		e.preventDefault();
 		e.stopPropagation();
-		window.open(`https://google.com/search?q=${task.user.name}`, '_blank', 'noreferrer');
+		window.open(`https://google.com/search?q=${entity.user.label}`, '_blank', 'noreferrer');
 	};
 	return (
 		<div style={{ cursor: 'pointer' }} onClick={handleClick}>
--- a/src/packages/admin-ui-components/src/detail-panel/component.tsx
+++ b/src/packages/admin-ui-components/src/detail-panel/component.tsx
@@ -163,23 +163,31 @@ export const DetailPanel = () => {
 		navigate(routeFor({ entity: selectedEntity, filters, sort }));
 	}, [search, selectedEntity]);
 
-	// Weed out ID fields - for the moment.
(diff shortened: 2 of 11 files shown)


src/packages/admin-ui-components/src/utils/use-schema.ts
```diff
@@ -32,13 +32,7 @@ export enum AdminUIFilterType {
...
 }
 
-export enum EntityFieldType {
-	BOOLEAN = 'Boolean',
-	ID = 'ID!',
-	OPTIONAL_ID = 'ID',
-	JSON = 'JSON',
-	// TODO: This needs to be extended for the other types
-}
+export type EntityFieldType = 'Boolean' | 'custom' | 'ID!' | 'ID' | 'JSON';
```
steve-keep, line 35: Curious why you removed the enum?

src/examples/rest/src/admin-ui/custom-fields/link.tsx
```diff
@@ -1,13 +1,20 @@
...
+		value: string;
+	};
+}
+
+export const Link = ({ entity }: CustomFieldArgs<Task>) => {
 	const handleClick = (e: MouseEventHandler<HTMLDivElement>) => {
 		e.preventDefault();
 		e.stopPropagation();
-		window.open(`https://google.com/search?q=${task.user.name}`, '_blank', 'noreferrer');
+		window.open(`https://google.com/search?q=${entity.user.label}`, '_blank', 'noreferrer');
```
steve-keep, line 17: Does the component get all the data on that row?

src/packages/admin-ui-components/src/utils/use-schema.ts
thekevinbrown: It wouldn't work from the user's custom fields definition. The import was undefined, so I just went to strings.

src/examples/rest/src/admin-ui/custom-fields/link.tsx
thekevinbrown: Yup. `entity` is the row object.

src/packages/admin-ui-components/src/detail-panel/component.tsx
```diff
@@ -181,6 +62,17 @@ const DetailField = ({ field }: { field: EntityField }) => {
 	);
 };
 
+const CustomField = ({ field }: { field: CustomField }) => {
```
steve-keep, line 65: Should this also be in the fields folder?

steve-keep (approved):
Looks good, just one small comment about moving the custom field into a file with the other fields.

src/packages/admin-ui-components/src/detail-panel/component.tsx
thekevinbrown: On this one, it's actually an alternative to `DetailField`, which is also in this file, e.g. the things in the folder are the actual fields, where these can more be called "wrappers"... What do you reckon?

steve-keep: sounds good 👍

PR #220, opened 2023-12-13
Remove turbo, revert to pnpm to watch build

tnodell:
Removes turbo and goes back to pnpm. Unclear why we used turbo.

Tested with two datasources and mikroORM example.

https://app.clickup.com/t/2704425/ID-5731


steve-keep (approved):

tnodell:
Closing to re-evaluate circular dependencies and look at watching with Turbo

Issue #222, opened 2023-12-14
Cannot determine GraphQL type for mutations with GraphQLJSON parameters


dylanlt:
**Describe the bug**
The admin ui fails to run with the latest package 0.2.9 if there are any mutations using the type GraphQLJSON. An example is the included mutation for `passkeyGenerateAuthenticationOptions`.

**To Reproduce**

1. Get 0.2.9, build and start
2. Navigate to Graphweaver admin ui

**Expected behavior**
The admin ui appears in the browser.

**Actual behavior**
HTTP error 502. In the terminal the error is `Cannot determine GraphQL output type for passkeyGenerateAuthenticationOptions`

**Screenshots or Screencasts**
<img width="1371" alt="image" src="https://github.com/exogee-technology/graphweaver/assets/7636222/1f7b2304-5bf1-4348-a855-12363f990749">

**Desktop (please complete the following information):**
- OS and version:  MacOS 14.1.2 M1

**Additional context**

steve-keep:
Thanks @dylanlt for raising the issue. 

It seems we had a mixture of `graphql` versions and a bug in `v16.6.0` where the custom scalars failed to register.

I have upgraded all instances to use `v16.8.1` (we had a mix of this version and the one above).

In your project, you may also want to ensure you are using `v16.8.1` and potentially use a pnpm override.

I will close this issue for now but if you continue to have issues please come back so we can help.

thanks again!

PR #223, opened 2023-12-15
GHI222: Scalar Bug

steve-keep:
This PR addresses the issue: https://github.com/exogee-technology/graphweaver/issues/222

To resolve we need to make sure that we are using v16.8.1 of GraphQL or higher as version v16.6.0 was causing the scalar types to become unregistered.

This also adds a test for the JSON scalar type to catch this in the future.


dylanlt (approved):
Great!

PR #224, opened 2023-12-15
Added Export to CSV

siowhoi:
Screenshot 2023-12-15 at 1 37 16 pm

### DIFF
--- a/src/packages/admin-ui-components/src/utils/export-csv.ts
+++ b/src/packages/admin-ui-components/src/utils/export-csv.ts
@@ -37,11 +37,11 @@ export const exportToCSV = <T>(entityname: string, columns: Column<T, unknown>[]
 	downloadFile(`${entityname} ${new Date().toString()}.csv`, blob);
 };
 
-function downloadFile(fileName: string, data: Blob) {
+const downloadFile = (fileName: string, data: Blob) => {
 	const downloadLink = document.createElement('a');
 	downloadLink.download = fileName;
 	const url = URL.createObjectURL(data);
 	downloadLink.href = url;
 	downloadLink.click();
 	URL.revokeObjectURL(url);
-}
+};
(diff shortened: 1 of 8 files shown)


src/packages/admin-ui-components/src/utils/export-csv.ts
```diff
@@ -0,0 +1,47 @@
+import Papa from 'papaparse';
+import type { Column } from 'react-data-grid';
+import { SelectOption } from '../multi-select';
+
+export const getExportDataFromSource = <T>(columns: Column<T, unknown>[], source: T[]) => {
+	const data =
+		source.map((row, index) =>
+			columns.map((c) => {
+				const value = row[c.key as keyof T] as any;
```
tnodell, line 9: Can we type this as at least having a label of type string? Is that possible?

tnodell (approved):
Looks ok to me, but have @steve-keep confirm that paparse is an ok library to use. And maybe double check we can't be more specific with types.

src/packages/admin-ui-components/src/utils/export-csv.ts
```diff
@@ -0,0 +1,47 @@
...
+			fields: columns.map((c) => c.name) as string[],
+			data,
+		},
+		{ delimiter: ',' }
+	);
+	const blob = new Blob([dataString], { type: 'text/csv;charset=utf-8' });
+	downloadFile(`${entityname} ${new Date().toString()}.csv`, blob);
+};
+
+function downloadFile(fileName: string, data: Blob) {
```
tnodell, line 40: const downloadFile = ()=> ? Usually we don't write `function` when we declare function

siowhoi: updated to use arrow function

src/packages/admin-ui-components/src/utils/export-csv.ts
siowhoi: As discussed with @steve-keep, all object values will be displayed as string in csv.
Screenshot 2023-12-19 at 3 30 19 pm

tnodell (approved):

Issue #183, opened 2023-10-12
Solution does not work


regasoraini:
**Describe the bug**
i follow the steps in the website : https://graphweaver.com/docs/quick-start
and i face this issue ( image attached )

**Screenshots or Screencasts**

`
× Unhandled exception in handler 'graphweaver-backend'.
× Runtime.ImportModuleError: Error: Cannot find module 'index'
  Require stack:
  - D:\Projects\Test Projects\test1\node_modules\.pnpm\serverless-offline@12.0.2_serverless@3.34.0\node_modules\serverless-offline\src\lambda\handler-runner\in-process-runner\aws-lambda-ric\UserFunction.js
      at _loadUserApp (D:\Projects\Test Projects\test1\node_modules\.pnpm\serverless-offline@12.0.2_serverless@3.34.0\node_modules\serverless-offline\src\lambda\handler-runner\in-process-runner\aws-lambda-ric\UserFunction.js:310:15)
      at async module.exports.load (D:\Projects\Test Projects\test1\node_modules\.pnpm\serverless-offline@12.0.2_serverless@3.34.0\node_modules\serverless-offline\src\lambda\handler-runner\in-process-runner\aws-lambda-ric\UserFunction.js:341:21)
      at async InProcessRunner.run (file:///D:/Projects/Test%20Projects/test1/node_modules/.pnpm/serverless-offline@12.0.2_serverless@3.34.0/node_modules/serverless-offline/src/lambda/handler-runner/in-process-runner/InProcessRunner.js:41:21)
      at async MessagePort.<anonymous> (file:///D:/Projects/Test%20Projects/test1/node_modules/.pnpm/serverless-offline@12.0.2_serverless@3.34.0/node_modules/serverless-offline/src/lambda/handler-runner/worker-thread-runner/workerThreadHelper.js:24:14)
× Error: Cannot find module 'index'
  Require stack:
  - D:\Projects\Test Projects\test1\node_modules\.pnpm\serverless-offline@12.0.2_serverless@3.34.0\node_modules\serverless-offline\src\lambda\handler-runner\in-process-runner\aws-lambda-ric\UserFunction.js

ANY / (λ: graphweaver-backend)
× Unhandled exception in handler 'graphweaver-backend'.
× Runtime.ImportModuleError: Error: Cannot find module 'index'
  Require stack:
[...]

PR #233, opened 2023-12-20
GHI: 183 Fix Build on Windows

steve-keep:
Graphweaver now has a test suite around the Windows OS and we are able to spin up an instance connected to Sqlite and test the connection.

This PR also addresses GHI https://github.com/exogee-technology/graphweaver/issues/183


src/packages/builder/src/start/backend.ts
```diff
@@ -19,9 +21,21 @@ import { config } from '@exogee/graphweaver-config';
 
 const rimraf = promisify(rimrafCallback);
 
+const isWindows = () => os.platform() === 'win32';
```
steve-keep, line 24: This is the change that fixed the windows bug.

.github/workflows/windows-end-to-end-tests.yml
```diff
@@ -0,0 +1,57 @@
+name: Windows End to End Tests
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  sqlite-end-to-end-tests:
+    runs-on: windows-latest
```
tnodell, line 9: 👍

tnodell (approved):

PR #247, opened 2024-01-03
Disable ID and readonly fields

tnodell:
https://app.clickup.com/t/2704425/ID-5905

This PR disables ID and readonly fields from the details panel.

When a readonly relationship is present, a new link field is used to link to the related entity. However, in the case that the detail panel form is dirty, these links do not work as we need to source a design for a confirmation modal.


src/packages/admin-ui-components/src/multi-select/component.tsx
```diff
@@ -32,7 +32,7 @@ export const Select = ({
 	mode = SelectMode.MULTI,
 }: SelectProps) => {
 	const [open, setOpen] = useState(false);
-	const [selectedOptions, setSelectedOptions] = useState<SelectOption[]>(value);
+	const [selectedOptions, setSelectedOptions] = useState<SelectOption[]>(value.flat());
```
tnodell, line 35: This is from https://github.com/exogee-technology/graphweaver/pull/244

src/packages/end-to-end/src/__tests__/ui/selectbox.test.ts
```diff
@@ -0,0 +1,11 @@
+import { test, expect } from '@playwright/test';
+
```
tnodell, line 2: So is this and the other test changes.
I needed this fix to see all the options

steve-keep (approved):

PR #253, opened 2024-01-04
Externals and bundled argon2

dylanlt:
https://app.clickup.com/t/2704425/ID-5960
A build error was introduced with argon2: 
`'No loader is configured for ".wasm" files: ../../node_modules/.pnpm/argon2-browser@1.18.0/node_modules/argon2-browser/dist/argon2.wasm'`

This fixes that error and declares some packages as externals to fix a different build error. A build test workflow has been added, too. This build test can be expanded later to include other packages such as admin-ui-components.


.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,46 @@
+name: Auth Build Tests
```
steve-keep, line 1: We can call this `Build Tests` and keep adding to it for any build-related tests.

.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  auth-build-tests:
```
steve-keep, line 8: same here drop the word `auth` and below in the step 👍

src/packages/auth/src/authentication/resolvers/password/resolver.ts
```diff
@@ -1,5 +1,5 @@
 import { BackendProvider, Resolver } from '@exogee/graphweaver';
-import * as argon2 from 'argon2-browser';
+import * as argon2 from 'argon2-browser/dist/argon2-bundled.min.js';
```
steve-keep, line 2: This import is unusual, can you add a comment above this line to explain why we are importing the file directly?

steve-keep (commented):
Overall this looks good, just a few tiny comments.

.github/workflows/build-tests.yml
dylanlt: Yes, missed that one!

src/packages/auth/src/authentication/resolvers/password/resolver.ts
dylanlt: Good pick-up. There was a lot of context here that is now captured.

steve-keep (approved):

Issue #250, opened 2024-01-03
Updating Album in SQLite DB errors


tnodell:
**Describe the bug**
When attempting to update an album using the SQLite example database in the end to end tests folder, an error message occurs

**To Reproduce**

1. Create an instance of the SQLite DB in the end to end test folder
2. Attempt to edit an album's name

**Expected behavior**
The album name is edited and saved

**Actual behavior**
An error message appears.

Error Message:
`Error: Variable "$data" got invalid value { id: "1", title: "For Those About To Rock We Salute You 2", artist: { id: "1" }, tracks: [[Object], [Object], [Object], [Object], [Object], [Object], [Object], [Object], [Object], [Object]] }; Field "tracks" is not defined by type "AlbumCreateOrUpdateInput".`

**Screenshots or Screencasts**
If applicable, add screenshots or screencasts to help explain the issue.

https://github.com/exogee-technology/graphweaver/assets/81122022/c2a43543-19be-4e1a-8224-6248a6ce82eb

**Desktop (please complete the following information):**

- OS and version:
Mac 14.1.1

**Additional context**
This also occurs when editing genre name, artist name, customer company.

agaw:
Thanks @tnodell, this issue has been fixed in https://github.com/exogee-technology/graphweaver/pull/255.

PR #255, opened 2024-01-05
Fix: Select component form value handling for OneToMany fields

agaw:
- Address Github issue 250: https://github.com/exogee-technology/graphweaver/issues/250
- Fixes an additional bug I noticed where changing the value of a multi-select field would result in only the last selected item being saved as the value for the multi-select field (and all existing selections would be deleted)
- Adds a unit test to some of the frontend logic and UI tests for the changes made in the points above
- Adds some metadata that helps VSCode recognise the workspace Typescript version


src/packages/admin-ui-components/src/detail-panel/util.ts
```diff
@@ -0,0 +1,19 @@
+export const processFormValues = (formValues: any) => {
+	const result = Object.entries(formValues ?? []).reduce((acc, [key, value]: [string, any]) => {
```
steve-keep, line 2: Maybe because it's a Friday afternoon 😄 But this code seems quite hard to follow. 

Now we have tests to cover its use could we rewrite it to make it easier to read?

I think maybe it's the reduce that could be replaced with a for-loop.

What do you think?

steve-keep (commented):
This looks good!

The only comment is on the `processFormValues` function. I found it a little difficult to follow and I wondered if we could make it simpler?

steve-keep (approved):

PR #256, opened 2024-01-08
Export Types

steve-keep:
This feature adds the ability for Graphweaver to output a types.ts file into ".graphweaver"  folder on each build.

You can specify the output of the types file by using the `typesDir ` cli option like this:

`graphweaver build-types --typesDir=./`
`graphweaver build --typesDir=./`
`graphweaver start --typesDir=./`
`graphweaver watch --typesDir=./`

This "types" file is a typescript definition file based on the GraphQL schema (SDL) output and can be used by front-end apps to get typed GraphQL queries and Mutations.


src/examples/mailchimp/src/backend/index.ts
```diff
@@ -15,10 +15,8 @@ if (!MAILCHIMP_LIST_ID) throw new Error('MAILCHIMP_LIST_ID is required in enviro
...
 
-const resolvers = [interest, category, mailingList, member];
-
-const graphweaver = new Graphweaver({
+export const graphweaver = new Graphweaver({
 	resolvers,
 });
 
-exports.handler = graphweaver.handler();
+export const handler = graphweaver.handler();
```
tnodell, line 22: We're exporting the graphweaver, handlers, and resolvers now. Are we importing them anywhere?

src/packages/builder/src/codegen/index.ts
```diff
@@ -56,9 +52,22 @@ export const codeGenerator = async () => {
...
-		);
+		// ensure that we have a graphweaver directory
+		const dirPath = path.join(process.cwd(), outdir || './.graphweaver');
+		if (!fs.existsSync(dirPath)) {
+			fs.mkdirSync(dirPath, { recursive: true });
+		}
+
+		const writeOperations = files.flatMap((file) => [
+			fs.promises.writeFile(path.join(process.cwd(), file.filename), file.content, 'utf8'),
+			// We save the types to two locations src and .graphweaver / outdir
```
tnodell, line 63: Why do we save the types to two locations?

src/packages/cli/bin/gw-types.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+/* eslint-disable @typescript-eslint/no-var-requires */
+var import_graphweaver_builder = require('@exogee/graphweaver-builder');
```
tnodell, line 3: Why is this file js

tnodell (approved):
Just a couple questions to ya.

src/packages/cli/bin/gw-types.js
steve-keep: The js file is a bin file and is executed in a mac or windows env. It is a `js` file to avoid any build step and follows the same pattern as the `gw-bin.js` file.

src/packages/builder/src/codegen/index.ts
steve-keep: The first location is used by the generated Apollo hooks and the types file is referenced from the other generated files.

The second location is for use outside of the app.

For example, you can specify the location to be inside another app you are working on. Im going to create a docs page on how to do this.

src/examples/mailchimp/src/backend/index.ts
steve-keep: No need to export the resolvers and yes we need to export both `graphweaver` and the `handler`.

The `handler` is imported by the lambda that runs the code.

The `graphweaver` instance is imported by the types generation to get the GraphQL schema.

Issue #248, opened 2024-01-03
Admin UI: Error displayed when navigating to a previously visited entity


agaw:
**Describe the bug**
The admin UI will sometimes show an unhandled error when navigating between entities.

**To Reproduce**

1. Follow the "Quick Start" guide to initialise a new Graphweaver project: [Docs Link](https://graphweaver.com/docs/quick-start)
2. Create a new, blank PostgreSQL database and populate it using the following SQL:
```
BEGIN;


CREATE TABLE address (
    id bigint NOT NULL,
    line_one character varying(255) NOT NULL,
    line_two character varying(255),
    city character varying(255),
    state character varying(255),
    postcode character varying(255),
    country character varying(255) NOT NULL
);

CREATE UNIQUE INDEX address_pkey ON address(id int8_ops);

CREATE TABLE campus (
    id bigint NOT NULL,
    name character varying(255) NOT NULL,
    address character varying(255),
    email character varying(255)
);

CREATE UNIQUE INDEX campus_pkey ON campus(id int8_ops);

CREATE TABLE student (
    id bigint NOT NULL,
    first_name character varying(255) NOT NULL,
    last_name character varying(255) NOT NULL,
    middle_name character varying(255),
    address_id bigint
);

CREATE UNIQUE INDEX student_pkey ON student(id int8_ops);
CREATE UNIQUE INDEX student_address_id_key ON student(address_id int8_ops);
CREATE UNIQUE INDEX student_address_id_unique ON student(address_id int8_ops);

INSERT INTO address VALUES (2, 'Some', 'Address', 'x', 'x', 'x', 'x');
INSERT INTO campus VALUES (1, 'Some Campus', 'Somewhere', 'x@x.com');
INSERT INTO student VALUES (1, 'John', 'Smith', 'x', 2);

COMMIT;
```
3. Follow the "Importing A Database" guide to create entity files for the database created in Step 2. [Docs Link](https://graphweaver.com/docs/importing-a-database)
4. Run `pnpm start` in the project root.
5. Navigate to `http://localhost:9000`, then click on the `mikro-orm-postgresql` database to reveal its tables.
6. Click on "Address", then "Campus", then "Student", then "Campus" again.

**Expected behavior**
The data in the "Campus" table is displayed.

**Actual behavior**
The following error message is displayed to the user in the browser:
```
Unhandled Thrown Error!
Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
[...]

PR #257, opened 2024-01-08
Fix React Hook order-related frontend crash

agaw:
This PR refactors out the use of a React hook in logic that is conditionally run. Under some specific circumstances (for example the recreation steps outlined in https://github.com/exogee-technology/graphweaver/issues/248) this can cause the UI to crash with a React hook order-related error.

### DIFF
--- a/src/packages/admin-ui-components/src/table/component.tsx
+++ b/src/packages/admin-ui-components/src/table/component.tsx
@@ -21,6 +21,10 @@ import { ApolloError } from '@apollo/client';
 
 import { customFields } from 'virtual:graphweaver-user-supplied-custom-fields';
 
+// Without stopping propagation on our links, the grid will be notified about the click,
+// which is not what we want. We want to navigate and not let the grid handle it
+const gobbleEvent = (e: MouseEvent<HTMLAnchorElement>) => e.stopPropagation();
+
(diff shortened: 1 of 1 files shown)


src/packages/admin-ui-components/src/table/component.tsx
```diff
@@ -38,10 +38,8 @@ const columnsForEntity = <T extends TableRowItem>(
 			? ({ row }: FormatterProps<T, unknown>) => {
 					// Without stopping propagation on our links, the grid will be notified about the click,
 					// which is not what we want. We want to navigate and not let the grid handle it
-					const gobbleEvent = useCallback(
-						(e: MouseEvent<HTMLAnchorElement>) => e.stopPropagation(),
-						[]
-					);
```
thekevinbrown, line 44: Instead of removing the `useCallback` optimisation, can we just move it out of the `map` so there's one instance for all the rows, e.g. put this on line 28?

thekevinbrown (changes requested):

src/packages/admin-ui-components/src/table/component.tsx
agaw: Thanks @thekevinbrown, I've moved it out completely to a single, global function.

thekevinbrown (approved):

PR #263, opened 2024-01-10
Types Feature Update

steve-keep:
This PR modifies the type feature in two ways:

1. The types output location is now done via config passed to Graphweaver and not as a cli flag
2. The default location for outputting the types file is `./src/frontend/types.ts`


agaw (approved):
Thanks @steve-keep, no comments on this one.

PR #266, opened 2024-01-15
Standalone End-To-End Test

agaw:
Implements https://app.clickup.com/t/2704425/ID-5971.

This PR adds a new Github Action that runs a 'standalone test' that checks that the build process results in a Lambda that can be run successfully. To do this, it creates a new Graphweaver project, imports a PostgreSQL database, then bundles the result; it then runs the resulting backend bundle using an emulated Lambda environment, and serves the frontend files using a simple http server. With this up and running, the existing backend and frontend tests are run.

Note the frontend tests have been modified slightly as some of the DOM selectors were different when running a Graphweaver instance using vite from the source directory vs the built/bundled static assets on a standalone server.


src/packages/end-to-end/src/__tests__/ui/issues/gh218.test.ts
```diff
@@ -13,16 +14,23 @@ test('Ensure filtering by multiple items works', async ({ page }) => {
...
 	// Select Accept in the artist filter
 	await page.getByRole('banner').getByText('AC/DC').click();
 	await page
-		.getByText('AC/DC×AC/DC×A Cor Do SomAC/DCAaron Copland & London Symphony OrchestraAaron Gold')
+		.locator('div')
+		.filter({ hasText: /^AC\/DC$/ })
+		.nth(1)
 		.click();
 
+	// TODO: Cannot currently test selecting multiple items in the filter because
```
steve-keep, line 28: Have we a ticket tracking this bug?

steve-keep (approved):
Looks good, just one comment about tracking the bug.

src/packages/end-to-end/src/__tests__/ui/issues/gh218.test.ts
agaw: We do now: https://app.clickup.com/t/86cudph7q

Issue #271, opened 2024-01-18
Build process hangs


dylanlt:
**Describe the bug**
After upgrading from 0.2.12 to 0.2.14 the build process hangs on `graphweaver build backend`

**To Reproduce**
Steps to reproduce the behaviour:

1. Create a mono-repo project 
2. Add a root build script that uses "turbo" or "pnpm -r", for example:
`"build-backend": "dotenv -- turbo run build-backend --only"`
3. Add a child package with a dependency on Graphweaver and a build script as follows:
`"build-backend": "graphweaver build backend"`
4. Run `pnpm build-backend` from the child folder
5. Run `pnpm build-backend` from the root folder

**Expected behaviour**
The build process runs and exits normally in both cases.

**Actual behaviour**
The build process hangs in the latter case.

**Desktop (please complete the following information):**

- OS and version: [MacOS 14.2.1 (M1 ARRM)

**Additional context**
Node runtimes 18 and 20 both have the same result. Logging `process._getActiveHandles()` shows that sockets have been left open.

dylanlt:
This has been resolved with version 0.2.16.

PR #272, opened 2024-01-19
Fix open handles

dylanlt:
https://github.com/exogee-technology/graphweaver/issues/271

This fixes an issue where the build can hang because of codegen.

Changes:

- explicitly stop the server after codegen has run
- switch from the cli to the core package of codegen


src/examples/rest/src/frontend/types.ts
```diff
@@ -21,179 +21,316 @@ export type Scalars = {
   JSON: { input: any; output: any; }
 };
 
-export type AdminUiEntityAttributeMetadata = {
-  __typename?: 'AdminUiEntityAttributeMetadata';
-  isReadOnly?: Maybe<Scalars['Boolean']['output']>;
+export type Query = {
```
thekevinbrown, line 24: I'd love to make this file `types.generated.ts` just so it matches what we're doing on the backend.

src/examples/xero/src/frontend/types.ts
```diff
@@ -21,31 +21,140 @@ export type Scalars = {
   ISOString: { input: any; output: any; }
 };
 
-export type Account = {
-  __typename?: 'Account';
-  code?: Maybe<Scalars['String']['output']>;
-  id: Scalars['ID']['output'];
-  name?: Maybe<Scalars['String']['output']>;
+export type Query = {
```
thekevinbrown, line 24: Same here, `types.generated.ts` would make it clear not to edit this file.

src/packages/builder/src/codegen/index.ts
```diff
@@ -1,114 +1,113 @@
...
+import * as typescript from '@graphql-codegen/typescript';
+import { loadDocuments } from '@graphql-tools/load';
+import { CodeFileLoader } from '@graphql-tools/code-file-loader';
+import { Source } from '@graphql-tools/utils';
 
 type CodegenOptions = {
 	typesOutputPath?: string | string[];
 };
 
+type CodegenConfig = { [key: string]: any };
```
thekevinbrown, line 17: I'd just inline `Record<string, any>` instead of declaring a type for this I think.

src/packages/builder/src/codegen/index.ts
```diff
@@ -1,114 +1,113 @@
...
+		documents,
+		filename: 'src/types.generated.ts',
+		config,
+		plugins: [{ typescriptPlugin: {} }],
+		pluginMap: {
+			typescriptPlugin: typescript,
+		},
+	});
+	fs.writeFileSync('src/types.generated.ts', content.concat('\n', typesFileContent));
+	fs.writeFileSync('src/frontend/types.ts', content.concat('\n', typesFileContent));
```
thekevinbrown, line 41:
```suggestion
	fs.writeFileSync('src/frontend/types.generated.ts', content.concat('\n', typesFileContent));
```

src/packages/builder/src/codegen/index.ts
```diff
@@ -1,114 +1,113 @@
...
+		},
+	};
+
+	let documents: Source[] | undefined = undefined;
+	try {
+		documents = await loadDocuments('./src/**/!(*.generated).{ts,tsx}', {
+			loaders: [new CodeFileLoader()],
+		});
+	} catch (e) {
+		documents = [];
```
thekevinbrown, line 95: What kind of errors is this hiding? If I run types generation and it doesn't do anything because of this catch, wouldn't it be nice to know with at least a log?

src/packages/cli/bin/gw-types.js
```diff
@@ -12,20 +12,19 @@ const generateTypes = async () => {
...
 	await builder.codeGenerator(sdl, codegenOptions);
+
+	await graphweaver.stopServer();
 };
 
 generateTypes()
-	.then(() => {
-		process.exit(0);
-	})
+	.then(() => {})
```
thekevinbrown, line 27: We can just remove this if it does nothing.

```suggestion
```

thekevinbrown (changes requested):
Some minor things to get to the bottom of, but looking good in general!

src/examples/rest/src/frontend/types.ts
```diff
@@ -21,179 +21,316 @@ export type Scalars = {
   JSON: { input: any; output: any; }
```
steve-keep, line 21: It looks like the output of the file is no longer alphabetical. Is it possible to get that ordering back?

src/packages/builder/src/codegen/index.ts
```diff
@@ -1,114 +1,113 @@
...
-		const writeOperations = files.flatMap((file) => [
-			fs.promises.writeFile(path.join(file.filename), file.content, 'utf8'),
-			// We save the types to two locations src and .graphweaver / outdir
-			...(file.filename === 'src/types.generated.ts'
-				? typesOutputPaths.map((outputPath: string) =>
-						fs.promises.writeFile(outputPath, file.content, 'utf8')
-				  )
-				: []),
-		]);
+	for (const args of argsList) {
```
steve-keep, line 65: Do we need this loop twice?

Looks like we are looping here and writing the file and then looping again on line 71?

src/packages/builder/src/codegen/index.ts
```diff
@@ -1,114 +1,113 @@
...
-	}
-	if (typesOutputPath && Array.isArray(typesOutputPath)) {
-		typesOutput.push(...typesOutputPath);
+	const config = {
+		skipDocumentsValidation: {
+			skipDuplicateValidation: true, // A flag to disable the validation for duplicate query and mutation names we don't need this as we are using near-operation-file
+		},
+	};
+
+	let documents: Source[] | undefined = undefined;
```
steve-keep, line 89: You could `const documents: Source[] = [];`

Then spread the result of `loadDocuments` into the array.

src/examples/rest/src/frontend/types.ts
dylanlt: I was just keeping the same name, but I had the exact same question. I'll change it.

src/packages/builder/src/codegen/index.ts
dylanlt: Correct - deleted

src/packages/builder/src/codegen/index.ts
dylanlt: Done

src/packages/builder/src/codegen/index.ts
dylanlt: This code is actually from the CLI; the common error is that none of the files matching this pattern are relevant to codegen: `Unable to find any GraphQL type definitions for the following pointers`

I will add a log here.

src/packages/builder/src/codegen/index.ts
dylanlt: Simplification applied

src/packages/builder/src/codegen/index.ts
dylanlt: Logging added.

src/packages/builder/src/codegen/index.ts
dylanlt: Simplification applied

src/examples/rest/src/frontend/types.ts
dylanlt: I don't know but I agree that sorting would be preferable. I will see what options there are.

dylanlt: Added "sort: true" to config

dylanlt:
Superseded by a simpler resolution in version 0.2.16

PR #286, opened 2024-01-23
Security Fix: Vite Upgrade

steve-keep:
This PR addresses the security issues raised by Dependabot.


src/packages/vite-plugin-graphweaver/package.json
```diff
@@ -22,7 +22,7 @@
 		"esbuild": "0.15.5",
 		"glob": "10.3.10",
 		"typescript": "5.0.2",
-		"vite": "4.3.9"
+		"vite": "4.5.2"
 	},
 	"peerDependencies": {
 		"vite": "4"
```
tnodell, line 28: I assume this doesn't need to specify a minor verision.

tnodell (approved):
Just checking that that peer dependency doesn't need to be bumped either.

src/packages/vite-plugin-graphweaver/package.json
steve-keep: This is fine for this PR but we may want to add vite 5 soon. Good catch though 👍

PR #299, opened 2024-02-01
Generate typed queries and mutations on clients that are outside the Graphweaver folder 

siowhoi:
Configure the Graphweaver to watch for file changes:

```
/* graphql-api Graphweaver Project */
import 'reflect-metadata';
import Graphweaver from '@exogee/graphweaver-server';
import { resolvers } from './schema';

export const graphweaver = new Graphweaver({
  resolvers,
  fileAutoGenerationOptions: {
    watchForFileChangesInPaths: ['../../apps/web-app'],
  },
});

export const handler = graphweaver.handler();
```
Example: https://github.com/siowhoi/graphweaver-example-monorepo


src/packages/builder/src/codegen/index.ts
```diff
@@ -14,13 +15,20 @@ const content = `/* eslint-disable */
...
+
 	try {
 		const files = await executeCodegen({
 			cwd: process.cwd(),
 			pluginLoader: async (plugin: string) => import(plugin),
 			schema,
 			ignoreNoDocuments: true,
-			documents: ['./src/**/!(*.generated).{ts,tsx}'],
+			documents: [
+				path.join('./src', extensions),
```
tnodell, line 27: Does this generate the types in both
`path.join('./src', extensions),`
and if provided in these paths?

```
watchForFileChangesInPaths?.map((clientPath) =>
					path.join(clientPath, extensions)
				) || []),
				```

tnodell (approved):
LGTM

src/packages/builder/src/codegen/index.ts
siowhoi: I have just tested if the same path appears in both and it works. :)

PR #302, opened 2024-02-06
Added support to display string/enum array as multi select dropdown in AdminUI

siowhoi:
1. Role in APIKey has been updated to enum array
2. Role will be displayed as multi select dropdown in AdminUI
3. Added unit test to assign role to an API key.

Screenshot 2024-02-06 at 11 25 28 am

### DIFF
--- a/src/examples/rest/database.sql
+++ b/src/examples/rest/database.sql
@@ -44,7 +44,7 @@ CREATE TABLE api_key (
   api_key VARCHAR(255) NOT NULL UNIQUE,
   secret VARCHAR(255) NOT NULL,
   revoked BOOLEAN NOT NULL DEFAULT false,
-  roles JSON NOT NULL DEFAULT (JSON_ARRAY())
+  roles TEXT NOT NULL
 );
 
 CREATE TABLE authentication (
--- a/src/examples/rest/src/backend/entities/mysql/api-key.ts
+++ b/src/examples/rest/src/backend/entities/mysql/api-key.ts
@@ -1,4 +1,4 @@
-import { BigIntType, Enum, Entity, PrimaryKey, Property } from '@mikro-orm/core';
+import { BigIntType, Enum, EnumArrayType, Entity, PrimaryKey, Property } from '@mikro-orm/core';
 import { BaseEntity } from '@exogee/graphweaver-mikroorm';
 import { ApiKeyStorage } from '@exogee/graphweaver-auth';
 import { Roles } from '../../auth/roles';
@@ -17,6 +17,6 @@ export class ApiKey extends BaseEntity implements ApiKeyStorage {
 	@Property({ type: Boolean, default: false })
 	revoked!: boolean;
 
-	@Enum({ type: 'enumArray', items: () => Roles, array: true, default: [] })
+	@Enum({ type: EnumArrayType, items: () => Roles, array: true, default: [] })
 	roles!: Roles[];
 }
--- a/src/examples/rest/src/backend/entities/mysql/api-key.ts
+++ b/src/examples/rest/src/backend/entities/mysql/api-key.ts
@@ -1,6 +1,7 @@
-import { ArrayType, BigIntType, Entity, PrimaryKey, Property } from '@mikro-orm/core';
+import { BigIntType, Enum, Entity, PrimaryKey, Property } from '@mikro-orm/core';
 import { BaseEntity } from '@exogee/graphweaver-mikroorm';
 import { ApiKeyStorage } from '@exogee/graphweaver-auth';
+import { Roles } from '../../auth/roles';
 
 @Entity({ tableName: 'api_key' })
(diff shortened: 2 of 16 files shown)


src/examples/rest/database.sql
```diff
@@ -44,7 +44,7 @@ CREATE TABLE api_key (
   api_key VARCHAR(255) NOT NULL UNIQUE,
   secret VARCHAR(255) NOT NULL,
   revoked BOOLEAN NOT NULL DEFAULT false,
-  roles JSON NOT NULL DEFAULT (JSON_ARRAY())
+  roles TEXT NOT NULL
```
tnodell, line 47: Is this change necessary? I would expect our docs and examples to break if we changed what kind of thing roles are. @steve-keep

steve-keep: MySQL does not have array support so there are two options TEXT or JSON.

I prefer JSON as you know that it is validated when saved to the DB but it is not perfect.

@siowhoi I assume you had to change this because the JSON type was not compatible with the @Enum decorator?

src/examples/rest/src/backend/entities/mysql/api-key.ts
```diff
@@ -16,6 +17,6 @@ export class ApiKey extends BaseEntity implements ApiKeyStorage {
 	@Property({ type: Boolean, default: false })
 	revoked!: boolean;
 
-	@Property({ type: ArrayType, default: [] })
-	roles!: string[];
+	@Enum({ type: 'enumArray', items: () => Roles, array: true, default: [] })
```
steve-keep, line 20: Do you need to include `array:true` here?

src/packages/server/src/metadata-service/resolver.ts
```diff
@@ -92,13 +92,13 @@ export const getAdminUiMetadataResolver = (hooks?: AdminMetadata['hooks']) => {
...
 						const typeValue = field.getType() as any;
-						const typeName = typeValue.name ? typeValue.name : enumMetadata.get(typeValue)?.name;
+						const typeName = typeValue.name ?? enumMetadata.get(typeValue)?.name;
 
 						const relatedObject = objectTypeData[typeName];
 
 						const fieldObject: AdminUiFieldMetadata = {
 							name: field.name,
-							type: relatedObject?.name || typeName,
+							type: relatedObject?.name || (field.typeOptions.array && `${typeName}[]`) || typeName,
```
steve-keep, line 101: I thought we decided to send the property `isArray` down to the client since the type did not have the string value `[]`? 

Do you prefer to concat the string like this?

src/packages/admin-ui-components/src/detail-panel/component.tsx
```diff
@@ -71,9 +71,11 @@ const getField = ({ field }: { field: EntityField }) => {
 	}
 
 	const { enumByName } = useSchema();
-	const enumField = enumByName(field.type);
+	const enumField = enumByName(field.type.replace('[]', ''));
 	if (enumField) {
-		return <EnumField name={field.name} typeEnum={enumField} />;
+		return (
+			<EnumField name={field.name} typeEnum={enumField} multiple={field.type.endsWith('[]')} />
```
steve-keep, line 77: Depending on whether we add the `isArray` flag here this might be `field.isArray`

src/packages/server/src/metadata-service/resolver.ts
siowhoi: Fixed. Added new property called `isArray`.

src/packages/admin-ui-components/src/detail-panel/component.tsx
siowhoi: Fixed. Added new property called `isArray`.

src/examples/rest/src/backend/entities/mysql/api-key.ts
siowhoi: Yes, if i remove `array: true`, i get error: `Expected Iterable`
Screenshot 2024-02-07 at 10 40 48 am

src/examples/rest/database.sql
siowhoi: Yes, I get error: Invalid JSON text. To fix this error, I will need a customType to convert the values to valid JSON string.

src/packages/admin-ui-components/src/detail-panel/component.tsx
```diff
@@ -71,9 +71,9 @@ const getField = ({ field }: { field: EntityField }) => {
 	}
 
 	const { enumByName } = useSchema();
-	const enumField = enumByName(field.type);
+	const enumField = enumByName(field.type.replace('[]', ''));
```
steve-keep, line 74: is this replace still needed?

siowhoi: Thanks for pointing it out. I have cleanup the code.

PR #312, opened 2024-02-19
Admin UI: Default Filter

steve-keep:
Adds a developer configurable default filter to the AdminUI for table views.

https://exogee.atlassian.net/jira/software/c/projects/EXOGW/boards/4?selectedIssue=EXOGW-313

### DIFF
--- a/src/packages/server/src/metadata-service/entity.ts
+++ b/src/packages/server/src/metadata-service/entity.ts
@@ -1,4 +1,7 @@
 import { Field, ObjectType } from 'type-graphql';
+import { GraphQLJSON } from '@exogee/graphweaver-scalars';
+import { Filter } from '@exogee/graphweaver';
+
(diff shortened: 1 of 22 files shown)


src/packages/server/src/metadata-service/entity.ts
```diff
@@ -16,6 +19,9 @@ export class AdminUiEntityMetadata {
 	@Field(() => [AdminUiFieldMetadata])
 	fields?: AdminUiFieldMetadata[] = [];
 
+	@Field(() => GraphQLJSON, { nullable: true })
+	defaultFilter?: Filter<any>;
```
thekevinbrown, line 23: Would `Filter<unknown>` be more accurate here?

thekevinbrown (approved):
Small Typescript question, other than that looks great!

Issue #268, opened 2024-01-15
Validation on required fields should occur before submitting


dylanlt:
**Describe the bug**
Required fields do not get validated in the UI before submitting.

**To Reproduce**

1. add {enableValidationRules: true} to Graphweaver option
2. pnpm i class-validator
3. add the @IsNotEmpty() decorator from class-validator to a field in an entity; this is now the required field
4. pnpm build & pnpm start
5. navigate to Graphweaver admin-ui and select this entity
6. click on the "create new" button at the top right
7. try to add a new instance of this entity but leave the required field empty

**Expected behavior**
When submit is clicked, a validation error is shown on the form and the form does not submit.

**Actual behavior**
When submit is clicked, the form submits and an. unfriendly backend error message is shown in a toast notification.

**Screenshots or Screencasts**
Example of an unfriendly message:

<img width="1440" alt="image" src="https://github.com/exogee-technology/graphweaver/assets/7636222/dbde60ff-fc93-4c5d-a34e-48d04dcf0734">

**Desktop (please complete the following information):**

- OS and version: Graphweaver 0.2.14 on MacOS 14.2.1 (M1 ARM)

**Additional context**
An idea is to potentially do this: https://medium.com/@serifcolakel/building-a-robust-unsaved-changes-prompt-with-react-and-react-router-dom-24f9157307ca

steve-keep:
We have now implemented client-side required validation in the AdminUI which can be used to get a better error message. For your use case, you may want to change the `ID` field to a sequence primary key and have a separate field to store your short code.

We treat `ID` as a special field on the form and assume that the ID is generated server side.

PR #315, opened 2024-02-21
AdminUI: Required Field Validation.

steve-keep:
If a field is not marked as nullable then it is considered required by the admin UI.

This will then be validated on the front end before submitting the form.

This PR partly addresses the issue raised in [GHI268](https://github.com/exogee-technology/graphweaver/issues/268).

Screenshot 2024-02-21 at 2 28 07 pm

### DIFF
--- a/src/packages/core/src/decorators/relationship-field.ts
+++ b/src/packages/core/src/decorators/relationship-field.ts
@@ -36,7 +36,7 @@ export function RelationshipField<
 	D extends BaseDataEntity = G['dataEntity']
 >(
 	returnTypeFunc: ReturnTypeFunc,
-	{ relatedField, id, nullable = true }: RelationshipFieldOptions<D>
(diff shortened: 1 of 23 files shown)


src/packages/core/src/decorators/relationship-field.ts
```diff
@@ -36,7 +36,7 @@ export function RelationshipField<
 	D extends BaseDataEntity = G['dataEntity']
 >(
 	returnTypeFunc: ReturnTypeFunc,
-	{ relatedField, id, nullable = true }: RelationshipFieldOptions<D>
+	{ relatedField, id, nullable = false }: RelationshipFieldOptions<D>
```
steve-keep, line 39: This is the correct change as it brings the default inline with the field resolver. However, it is potentially a breaking change.

thekevinbrown (approved):

PR #318, opened 2024-02-23
Error Fallback Screen

thekevinbrown:
It is possible that
1. A developer builds a custom field or page that throws an error.
2. Graphweaver has bugs in some scenarios which generate errors.

This PR adds a fallback error component which looks like this:

Screenshot 2024-02-26 at 11 24 20 am

* It `console.error()`s the error so the developer can see the full error in the console.
* It also gives you a button to copy error information to the clipboard.


src/packages/admin-ui-components/src/default-error-fallback/component.tsx
```diff
@@ -0,0 +1,78 @@
...
+				<Spacer height={10} />
+				<p>
+					If you are not a developer, please contact the person or company that set Graphweaver up
+					for you and tell them about this error.
+				</p>
+				<Spacer height={10} />
+				<p>
+					If you are a developer, this is likely a problem in a custom field or page. It could also
+					be a problem in Graphweaver itself. If you believe this is a Graphweaver bug, please{' '}
+					<a href="" target="_blank">
```
steve-keep, line 46: Did you want to include the uri here?

steve-keep (approved):

src/packages/admin-ui-components/src/default-error-fallback/component.tsx
thekevinbrown: Yes I do indeed. Done!

PR #321, opened 2024-02-27
Example App: SQLite

steve-keep:
This PR adds a detailed SQLite example to the repo.


src/examples/sqlite/package.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "dependencies": {
+        "@as-integrations/aws-lambda": "2.0.1",
+        "@exogee/graphweaver": "workspace:*",
+        "@exogee/graphweaver-scalars": "workspace:*",
+        "@exogee/graphweaver-server": "workspace:*",
+        "@exogee/graphweaver-mikroorm": "workspace:*",
+        "@mikro-orm/core": "6.1.5",
+        "@mikro-orm/knex": "6.1.5",
+        "@mikro-orm/sqlite": "6.1.5",
```
thekevinbrown, line 22: Do we actually need this given that we have `mikro-orm-sqlite-wasm`?

src/examples/sqlite/src/backend/schema/invoice/entity.ts
```diff
@@ -0,0 +1,40 @@
...
+	billingState?: string;
+
+	@Field(() => String, { nullable: true })
+	billingCountry?: string;
+
+	@Field(() => String, { nullable: true })
+	billingPostalCode?: string;
+
+	@Field(() => String)
+	total!: string;
```
thekevinbrown, line 36: Might be worth a comment here about why this is a string instead of a number, or we can just make it a number for going to and from the wire.

src/examples/sqlite/src/backend/schema/track/entity.ts
```diff
@@ -0,0 +1,53 @@
...
+	composer?: string;
+
+	@Field(() => Number)
+	milliseconds!: number;
+
+	@Field(() => Number, { nullable: true })
+	bytes?: number;
+
+	@Field(() => String)
+	unitPrice!: string;
```
thekevinbrown, line 46: Same here, we should probably put a comment here about why prices are in strings or change it so it's a number type.

src/examples/sqlite/src/backend/entities/sqlite/invoice-line.ts
```diff
@@ -0,0 +1,32 @@
...
+	@ManyToOne({
+		entity: () => Track,
+		ref: true,
+		fieldName: 'TrackId',
+		index: 'IFK_InvoiceLineTrackId',
+	})
+	track!: Ref<Track>;
+
+	@Property({ fieldName: 'UnitPrice', type: 'NUMERIC(10,2)' })
+	unitPrice!: string;
```
thekevinbrown, line 28: Same here, should be a comment about why this is a string I reckon.

src/examples/sqlite/src/backend/entities/sqlite/invoice.ts
```diff
@@ -0,0 +1,50 @@
...
+	billingState?: string;
+
+	@Property({ fieldName: 'BillingCountry', type: 'NVARCHAR(40)', nullable: true })
+	billingCountry?: string;
+
+	@Property({ fieldName: 'BillingPostalCode', type: 'NVARCHAR(10)', nullable: true })
+	billingPostalCode?: string;
+
+	@Property({ fieldName: 'Total', type: 'NUMERIC(10,2)' })
+	total!: string;
```
thekevinbrown, line 46: Comment would be good.

thekevinbrown (approved):
Looks great, just think we have an extra dep we don't need and some comments around why `NUMERIC` becomes `string` would be useful to people looking over this example.

src/examples/sqlite/package.json
steve-keep: yeah we do `mikro-orm-sqlite-wasm` uses it 👍

Issue #297, opened 2024-01-31
Table named 'fish' (among other options that do not pluralise easily) prevents Graphweaver from generating certain CRUD operations


mackcheesman:
**Describe the bug**
Due to quirks in the english language, some naming choices in the schemas of my data providers prevent Graphweaver from creating CRUD operations for multiple entities of a given type at once. 

For example, it creates operations such as `createFish, updateFish, deleteFish` and `fish` to operate over a single `Fish` entity, but since one fish is a 'fish', and two or more fish are multiple 'fish', equivalents aren't made for operating on multiple fish.

**To Reproduce**

1. Create a database with a table in it called 'fish' (internal schema is not particularly relevant)
2. Use `graphweaver import` to generate backend logic
3. Run `pnpm build && pnpm start`
4. Access the admin UI
5. Under the data provider, click on the `Fish` table.

**Expected behavior**
As with any other table in my data provider, I should be able to see these entities.

**Actual behavior**
The admin UI produces a 400 error. Using the introspection query reveals that some operations are missing.

**Screenshots or Screencasts**
<img width="1231" alt="Screenshot 2024-01-31 at 10 58 13 am" src="https://github.com/exogee-technology/graphweaver/assets/112445159/caa44d15-5aef-4ec0-8e81-1040b229f89b">

**Desktop (please complete the following information):**

- OS and version: MacOS 14.2.1 (M1)

thekevinbrown:
In the backlog we plan to have a feature to handle this consistently. We should either implement that, or at least make a better error message for developers when `plural(tableName) === tableName`.

thekevinbrown:
The approach we settled on in this PR ^:

* Use the existing `pluralize()` function by default.
* If `pluralize(entityName) === entityName` then use `'multiple'` for the plural instead, e.g. `createFish`, `createMultipleFish`, etc.
* Developers can pass in an option called `plural` like so:
```typescript
@Resolver((of) => Fish)
class FishResolver extends createBaseResolver<Fish, any>(Fish, new BaseDataProvider('fish'), {
	plural: 'fishes',
}) {}
```
* When that option is passed in, we use it instead of the above, e.g. `createFish`, `createFishes`, etc.
* At all times if a collision is detected, we throw an error and refuse to start up so the developer knows what's going on.

PR #335, opened 2024-02-29
Support Plural Names 🐠

steve-keep:
This PR addresses the Fish situation that is detailed here:

https://github.com/exogee-technology/graphweaver/issues/297

There is also a refactor of the metadata store and entity map.


.vscode/launch.json
```diff
@@ -43,7 +43,7 @@
       "sourceMaps": true,
       "cwd": "${workspaceFolder}/src/examples/mailchimp"
     },
-    {
+     {
```
thekevinbrown, line 46: Extremely pedantic, I know:
```suggestion
    {
```

src/packages/end-to-end/package.json
```diff
@@ -16,11 +16,12 @@
 		"dev-mysql": "pnpm import-database-mysql && pnpm start-server",
 		"unpack": "tsx ./scripts/unpack.ts",
 		"test-auth": "jest --runInBand --forceExit -- api/auth",
+		"test-aws-cognito": "jest --runInBand --forceExit ---- api/aws-cognito",
+		"test-metadata": "jest --runInBand --forceExit ---- api/metadata",
 		"test-mysql": "export DATABASE=mysql && jest --runInBand -- api/mysql",
 		"test-postgres": "export DATABASE=postgres && jest --runInBand -- api/postgres",
 		"test-sqlite": "dotenv -e .env.sqlite -- jest --runInBand --forceExit ---- api/sqlite",
 		"test-scalars": "jest --runInBand --forceExit ---- api/scalars",
-		"test-aws-cognito": "jest --runInBand --forceExit ---- api/aws-cognito",
```
thekevinbrown, line 23: Should we add a new `test-api` that just does:
```
"test-api": "jest --runInBand --forceExit ---- api",
```
and use that in CI, just so that if someone adds a new one but doesn't hook it all the way up like this it doesn't accidentally not run?

thekevinbrown (approved):
Looks awesome, thanks!

.vscode/launch.json
steve-keep: Hmm thats annoying let me tweak my vscode settings.

src/packages/end-to-end/package.json
steve-keep: Not a bad shout but many of the API tests rely on an end to end app. So adding the script would not work out the box.

Maybe we can add a new ticket to tidy up the pattern for test runs?

thekevinbrown: Yup, sounds good!

PR #336, opened 2024-03-01
AdminUI: Simplify Filter Object

steve-keep:
This is prerequisite work to get ready for the more complex filter component in the Admin UI. The filter structure is now the same as the Filter used by the GraphQL API.


src/packages/admin-ui-components/src/filter-bar/component.tsx
```diff
@@ -1,129 +1,136 @@
...
+	const filterFieldsOnEntity = useCallback(() => {
+		const entity = entityByName(entityName);
 		// @todo - currently the filters are not fitting on the screen
 		// we plan to redo this filter bar so that it is a drop down
 		// for now the workaround is to reduce the number of filters to 5
+		const showOnlyFiveFilters = entity.fields.length > 5 ? 5 : entity.fields.length;
+		const fields = entity.fields
+			// filter out rowEntity.fields with the JSON type
+			.filter((field) => field.type !== 'JSON')
+			.slice(0, showOnlyFiveFilters);
```
thekevinbrown, line 40: Couldn't we simplify this to:
```typescript
const fields = entity.fields
	// filter out rowEntity.fields with the JSON type
	.filter((field) => field.type !== 'JSON')
	.slice(0, 5);
```
and just get rid of line 36? If not, it'd be like this I reckon:
```typescript
const fields = entity.fields
	// filter out rowEntity.fields with the JSON type
	.filter((field) => field.type !== 'JSON')
	.slice(0, Math.min(5, entity.fields.length));
```

thekevinbrown (approved):
Looks great!

src/packages/admin-ui-components/src/filter-bar/component.tsx
steve-keep: first one is good 👍

PR #337, opened 2024-03-04
Apollo Client Package

steve-keep:
We have some useful client side Apollo helpers that are useful to not only AdminUI but to any app intergrating with Graphweaver.

This PR moves the first helpful function (`generateTypePolices`) out into a new package that can be shared amongst apps.

### DIFF
--- a/src/packages/apollo-client/src/index.ts
+++ b/src/packages/apollo-client/src/index.ts
@@ -12,9 +12,11 @@ const generateTypePolicyFields = (entityNames: string[]) => {
 			return btoa(`${filter}:${orderBy}`);
 		},
 		merge(existing = [], incoming: { __ref: string }[]) {
-			const merged = [...existing, ...incoming];
-			const uniqueItems = new Set(merged.map((item) => item.__ref));
-			return [...uniqueItems].map((__ref) => merged.find((item) => item.__ref === __ref));
+			const mergeMap = new Map<string, { __ref: string }>();
+			for (const entity of [...existing, ...incoming]) {
+				mergeMap.set(entity.__ref, entity);
+			}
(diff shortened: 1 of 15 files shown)


src/packages/apollo-client/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+		) => {
+			// https://www.apollographql.com/docs/react/pagination/key-args/#keyargs-function-advanced
+			const filter = args?.filter ? JSON.stringify(args.filter) : '';
+			const orderBy = args?.pagination?.orderBy ? JSON.stringify(args.pagination.orderBy) : '';
+			return btoa(`${filter}:${orderBy}`);
+		},
+		merge(existing = [], incoming: { __ref: string }[]) {
+			const merged = [...existing, ...incoming];
+			const uniqueItems = new Set(merged.map((item) => item.__ref));
+			return [...uniqueItems].map((__ref) => merged.find((item) => item.__ref === __ref));
```
thekevinbrown, line 17: So currently we're:

1. Creating a new array of everything in it
2. For looping that array to make a Set of refs
3. Making a new array of those refs, then for each of those, for looping the array in step 1 to look for it.

Couldn't we do this instead:
```typescript
const merged = new Map<Ref, Entity>();
for (const entity of [...incoming, ...existing]) {
    merged.set(entity.__ref, entity);
}
return merged.values();
```

thekevinbrown (approved):
I think we can be more efficient in our `merge` algorithm, but I'll approve because that's currently working.

src/packages/apollo-client/src/index.ts
steve-keep: Updated this is much better!

PR #342, opened 2024-03-11
FIX: Many Deletes Failed

steve-keep:
This is a fix to ensure that deletes work for Mikro.

The existing delete test was passing even when throwing an error. I have removed this and replaced with a new test in the auth ui package (which uses the REST example).

The reason I had to use the REST example for this test was because:

- SQLite is not allowing the deletion of data
- There is "NO ACTION" on deletes in the chinook dataset and there are FK on all entities so not possible to delete data
- There are no AUTO_INCREMENT IDs in the postgres dataset so not currently able to create data in the tests

There is now https://exogee.atlassian.net/browse/EXOGW-340 to track the work of adding mutation tests for these datasets.


src/examples/rest/src/backend/schema/tag/entity.ts
```diff
@@ -29,7 +29,7 @@ const acl: AccessControlList<Tag, AuthorizationContext> = {
 };
 
 @ApplyMultiFactorAuthentication<Tag>(() => ({
-	Everyone: {
+	LIGHT_SIDE: {
```
steve-keep, line 32: In order to test delete I have changed this MFA to only trigger for Light Side role.

thekevinbrown (approved):

PR #344, opened 2024-03-12
Introspection: Check for unique foreign keys

steve-keep:
Introspection will fail if the database has duplicate foreign keys.

This PR checks for any duplicates and alerts the dev to the table and column where the issue was found:

Screenshot 2024-03-12 at 2 28 04 pm

### DIFF
--- a/src/packages/mikroorm/src/introspection/generate.ts
+++ b/src/packages/mikroorm/src/introspection/generate.ts
@@ -141,9 +141,9 @@ const assertUniqueForeignKeys = (table: DatabaseTable): void => {
 		const { localTableName, referencedTableName, columnNames, referencedColumnNames } = foreignKey;
 		const serializedValue = JSON.stringify({
 			localTableName,
-			columnNames,
+			columnNames: columnNames.sort(),
 			referencedTableName,
-			referencedColumnNames,
+			referencedColumnNames: referencedColumnNames.sort(),
 		});
 
 		if (uniqueForeignKeys.has(serializedValue)) {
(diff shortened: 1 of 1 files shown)


src/packages/mikroorm/src/introspection/generate.ts
```diff
@@ -133,6 +133,31 @@ const convertToCamelCasePropertyNames = (metadata: EntityMetadata[]): void => {
 	}
 };
 
+const assertUniqueForeignKeys = (table: DatabaseTable): void => {
+	const uniqueForeignKeys = new Set();
+	const definedForeignKeys = Object.values(table.getForeignKeys());
+
+	for (const foreignKey of definedForeignKeys) {
+		const { localTableName, referencedTableName, columnNames, referencedColumnNames } = foreignKey;
+		const serializedValue = JSON.stringify({
```
thekevinbrown, line 142: `JSON.stringify` is deterministic as long as the properties were created in the same order on the object, which in this case they will be, so all good, just noting after doing a bit of digging.

thekevinbrown (approved):

src/packages/mikroorm/src/introspection/generate.ts
steve-keep: yeah I had the same thought....

Is there a easy way or do we literally need to sort it first for each array?

steve-keep: they are just string arrays so we should be ok with adding just sort.

PR #649, opened 2024-05-14
Exogw 342

steve-keep:
This is a major update that removes the dependency of Type GraphQL.


steve-keep:
@coderabbitai review

src/packages/core/src/resolvers.ts
```diff
@@ -0,0 +1,475 @@
...
+		| GraphQLEntityConstructor<G, D>[];
+	let isList = false;
+
+	if (Array.isArray(gqlEntityType)) {
+		isList = true;
+		gqlEntityType = gqlEntityType[0];
+	}
+
+	// @todo: Should the user specified filter be and-ed here?
+	//        My worry is if we just pass the filter through, it could be used to circumvent the relationship join.
```
steve-keep, line 393: Should we start by anding? That way we know our filter is always applied, I think that is how I assumed it would work when using it as you can potentially get data that is not related to the parent.

thekevinbrown: Yeah I'm down with that.

steve-keep: done 🚀

src/packages/storage-provider/src/storageProvider.ts
```diff
@@ -0,0 +1,150 @@
...
+		});
+	}
+
+	async getUploadUrl(
+		_source: Source,
+		{ key }: { key: string },
+		_ctx: BaseContext,
+		_info: GraphQLResolveInfo
+	): Promise<{ url: string; filename: string; type: MediaType }> {
+		if (!this.bucketName) throw new Error('Missing required env AWS_S3_BUCKET');
```
steve-keep, line 82: There are no hook calls or ACL checks on these mutations and query.

Should we add them as part of this work or create a new task to track it?

thekevinbrown: Let's go ahead and finish this properly as part of this work.

src/packages/aws-cognito/src/util.ts
```diff
@@ -47,7 +47,7 @@ export const createUser = async (
 	if (!user) return null;
 	return {
 		...user,
-		Groups: groups.split(','),
+		Groups: groups?.split(','),
```
thekevinbrown, line 50: Might be a bit nicer to do:
```suggestion
		Groups: groups?.split(',') ?? [],
```

src/packages/core/src/schema-builder.ts
```diff
@@ -0,0 +1,832 @@
...
+		});
+
+		updateTypes.set(entity.name, updateType);
+	}
+
+	return updateType;
+};
+
+export interface SchemaBuilderOptions {
+	authChecker?: AuthChecker<any, any>;
```
thekevinbrown, line 510: I think this is a hangover from TypeGraphQL. Is this actually getting called anywhere now?

src/packages/core/src/types.ts
```diff
@@ -199,3 +199,62 @@ export interface BackendProviderConfig {
...
+	complexity?: Complexity;
+	defaultValue?: any;
+
+	// This marks the field as read only in both the API and the admin UI.
+	// This will supersede any other read only settings.
+	readonly?: boolean;
+
+	nullable?: boolean | 'items' | 'itemsAndList';
+	excludeFromFilterType?: boolean;
+	hideInAdminUI?: boolean;
```
thekevinbrown, line 249:
```suggestion
```
Users can use hideInTable and hideInFilterBar together if they wish.

src/packages/mikroorm/src/database.ts
```diff
@@ -1,4 +1,3 @@
-import 'reflect-metadata';
```
thekevinbrown, line 1: I think we might actually need this in the MikroORM package only.

src/packages/mikroorm/src/provider/provider.ts
```diff
@@ -363,7 +363,7 @@ export class MikroBackendProvider<D extends BaseDataEntity, G extends GraphQLEnt
 		filter?: any
 	): Promise<D[]> {
 		const queryFilter = {
-			$and: [{ [relatedField]: { $in: relatedFieldIds } }, ...[filter ?? []]],
+			$and: [{ [relatedField]: { $in: relatedFieldIds } }, ...[gqlToMikro(filter) ?? []]],
```
thekevinbrown, line 366: Good catch!

src/packages/storage-provider/src/decorators/media-field.ts
```diff
@@ -0,0 +1,134 @@
...
+	BaseDataEntity,
+	Entity,
+	Field,
+	FieldOptions,
+	GraphQLEntity,
+	graphweaverMetadata,
+	FieldMetadata,
+} from '@exogee/graphweaver';
+import { S3StorageProvider } from '../storageProvider';
+import { Source } from 'graphql';
```
thekevinbrown, line 11: We're exporting from `graphql/types` already from `@exogee/graphweaver`. Should we also export `Source` so people don't have to have their own `graphql` dependency to get in trouble?

src/packages/storage-provider/src/decorators/media-field.ts
```diff
@@ -0,0 +1,134 @@
...
+	static serialize = ({ value }: { value: unknown }) => {
+		if (value === null) return null;
+		if (isMedia(value)) {
+			return JSON.stringify(
+				{
+					filename: value.filename,
+					type: value.type,
+				},
+				null,
+				2
```
thekevinbrown, line 66: I reckon we should save wire space here and remove the `null, 2` parameters. I know the DB will then store it however it wants, so no need to make it pretty on the way through I reckon.

src/packages/storage-provider/src/storageProvider.ts
```diff
@@ -0,0 +1,150 @@
...
+			...(this.endpoint ? { endpoint: this.endpoint } : {}),
+		});
+
+		const fileExtension = key.split('.').pop();
+		if (!fileExtension) {
+			throw new Error('Invalid file extension');
+		}
+
+		// generate uuid from crypto module
+		const uuid = randomUUID();
```
thekevinbrown, line 95: Random thought, but if we use uuid v7, we'll get time info on the filenames too, which could be really handy: https://buildkite.com/blog/goodbye-integers-hello-uuids

thekevinbrown (approved):
Nothing worth blocking over, major feat!

src/examples/aws-cognito/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@exogee/graphweaver-example-aws-cognito",
-	"version": "0.2.24",
+	"version": "1.0.0-beta-.1",
```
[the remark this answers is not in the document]
steve-keep: Ah! I used `.` to follow semver spec:

https://semver.org/

src/examples/s3-storage/src/frontend/types.ts
```diff
@@ -17,8 +17,8 @@ export type Scalars = {
   Boolean: { input: boolean; output: boolean; }
```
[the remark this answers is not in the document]
steve-keep: done!

src/packages/core/src/types.ts
steve-keep: I cant see any refs to hideInAdminUI so just removed.

src/packages/mikroorm/src/database.ts
steve-keep: I think we are ok as there is a ref at the top of `src/packages/mikroorm/src/index.ts`

src/packages/auth/src/authentication/methods/magic-link.ts
```diff
@@ -0,0 +1,320 @@
...
+		if (!link) {
+			throw new AuthenticationError('Authentication Failed: Magic Link not found');
+		}
+
+		await this.provider.updateOne(id, {
+			data: {
+				...link.data,
+				redeemedAt: new Date(),
+			},
+		});
```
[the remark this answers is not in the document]
steve-keep: OK so I think there are two things:

1. Can a magic link be used twice
2. Consolidation of `redeemMagicLink` inside `verifyMagicLink`

For 1, we have this on `getMagicLink`:

```
const link = await this.provider.createOne({
			type: AuthenticationType.MagicLinkChallenge,
			userId,
			data: {
				token,
				redeemedAt: 'null',
			},
		});
```

So you cant use the same link twice once it has a date it wont come back.

Should we merge `redeemMagicLink` inside `verifyMagicLink`?

src/packages/storage-provider/src/decorators/media-field.ts
steve-keep: done 👍

src/packages/storage-provider/src/storageProvider.ts
steve-keep: Unless you are super keen I think v4 is ok is instance as we are only using it for a unique name. We could also add created timestamp to the filename if needed as it would be human readable?

src/examples/rest/src/backend/schema/tag.ts
```diff
@@ -35,17 +30,18 @@ const acl: AccessControlList<Tag, AuthorizationContext> = {
...
 	id!: string;
 
-	@SummaryField()
 	@Field(() => String)
 	name!: string;
 
-	@RelationshipField<Task>(() => [Task], { relatedField: 'tags' })
-	tasks!: Task[];
+	@RelationshipField<Task>(() => [Task], { relatedField: 'tags', nullable: true })
+	tasks?: Task[];
```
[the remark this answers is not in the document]
steve-keep: This was more of an issue on how we deal with it on the Admin UI.

The field becomes required on both sides. So when you have a many-to-many and both are required you cant create an item 😄

src/examples/aws-cognito/package.json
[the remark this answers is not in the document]
steve-keep: As discussed wrong version of the spec and we should remove the trailing dash so from `1.0.0-beta-.1` to `1.0.0-beta.1`

src/packages/storage-provider/src/storageProvider.ts
steve-keep: As discussed as UUIDv7 is not part of crypto we will use UUIDv4 for now.

src/packages/auth/src/decorators/hooks/acl.ts
```diff
@@ -28,7 +22,46 @@ import {
 	isPopulatedFilter,
 } from '../../auth-utils';
 
-const metadata = getMetadataStorage();
+const metadata = graphweaverMetadata;
+
+const getAccessTypeForArg = (node: unknown): AccessType => {
```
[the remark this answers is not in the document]
steve-keep: Added comments to better explain the role of this function

src/packages/auth/src/authentication/methods/magic-link.ts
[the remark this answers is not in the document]
steve-keep: Consolidated the extra functions into the mutation functions.

src/packages/storage-provider/src/storageProvider.ts
steve-keep: Captured as part of https://exogee.atlassian.net/browse/EXOGW-355

src/examples/rest/src/backend/schema/tag.ts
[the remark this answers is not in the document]
steve-keep: https://exogee.atlassian.net/browse/EXOGW-71

Issue #775, opened 2024-06-20
Version mismatch on mikro-orm for sqlite


skorov:
**Describe the bug**
The initial project folder created by graphweaver has a version mismatch between mikro-orm and it's plugins. This causes an error when trying to import a sqlite database.

**To Reproduce**

1. Follow the instructions here <https://graphweaver.com/docs/quick-start>
2. Follow the instructions here <https://graphweaver.com/docs/connect-to-sqlite>
3. Should see the error while importing the database

**Expected behavior**
No error and the database imports successfully

**Actual behavior**
The following error occurs:

```
root@a88af515fadc:/volkis_test# npx graphweaver@latest import sqlite --database=Chinook_Sqlite.sqlite --host localhost
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
Importing data source...
Source: sqlite
Database Name: Chinook_Sqlite.sqlite
Database Host: localhost

(node:641) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
⠋ Introspecting...

Warning: Unable to connect to database.
Bad @mikro-orm/knex version 6.2.8.
All official @mikro-orm/* packages need to have the exact same version as @mikro-orm/core (6.2.9).
Only exceptions are packages that don't live in the 'mikro-orm' repository: nestjs, sql-highlighter, mongo-highlighter.
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
```

**Desktop:**

- OS and version: Debian 12 running Docker node@latest

**Additional context**
Seems like a bump to version 6.2.9 in the package.json template fixes the issue.

Specifically, the `MIKRO_ORM_TARGET_VERSION` variable in `/src/packages/cli/src/init/constants.ts`

thekevinbrown:
Thanks for reporting, we'll cut a release shortly!

thekevinbrown:
Version 1.0.1 is out, which we believe will fix this issue. If you hit any new ones just let us know!

PR #776, opened 2024-06-20
Fix / Mikro version mismatch

thekevinbrown:
We used Dependabot to update MikroORM without realising we hadn't updated the start template.

This PR:
- Updates the target version of MikroORM to fix the issue reported in #775
- Adds a build step to ensure they don't ever get out of sync again.

### DIFF
--- a/src/packages/cli/build.js
+++ b/src/packages/cli/build.js
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const { exec: nodeExec } = require('child_process');
+const path = require('node:path');
+const fs = require('node:fs/promises');
 const { promisify } = require('util');
 
 const { build } = require('esbuild');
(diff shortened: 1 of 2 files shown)


src/packages/cli/build.js
```diff
@@ -1,13 +1,51 @@
...
+	const devVersion = mikroDevDependencies['@mikro-orm/core'];
+	const peerVersion = mikroPeerDependencies['@mikro-orm/core'];
+	if (devVersion !== peerVersion) {
+		throw new Error(
+			`The version of @mikro-orm/core in the mikroorm package's devDependencies (${devVersion}) and peerDependencies (${peerVersion}) do not match.`
+		);
+	}
+
+	// Ok, what's ours?
+	const searchPattern = /export const MIKRO_ORM_TARGET_VERSION = '(.+)';/;
```
steve-keep, line 31: Should we move it to a config json rather than string match? My concern here is that someone breaks the string match by accident.

thekevinbrown: Yup, I'm down, just didn't want to involve `tsx` in the build process. I'll JSON-ify it.

thekevinbrown: Even better, I fixed it by just deriving it from the package.json file of the mikroorm package so there's no constant to update.

src/packages/cli/build.js
```diff
@@ -1,13 +1,51 @@
...
+	if (peerVersion !== MIKRO_ORM_TARGET_VERSION) {
+		throw new Error(
+			`The version of @mikro-orm/core in the mikroorm package's peerDependencies (${peerVersion}) does not match the target version in the CLI (${MIKRO_ORM_TARGET_VERSION}). Update the CLI package's src/init/constants.ts file to match the target version.`
+		);
+	}
+};
+
 (async () => {
+	await validateMikroOrmExpectedVersion();
+
```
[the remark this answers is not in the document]
thekevinbrown: Done

steve-keep (approved):

PR #799, opened 2024-06-27
Local Tracing

steve-keep:
Show OpenTelemetry Traces inside the Admin UI:

Screenshot 2024-07-02 at 4 31 30 pm


src/examples/sqlite/databases/database.sqlite
steve-keep, line 1: I wonder if this and the trace db should be removed from the examples and instructions added to generate them or a script that does it first time?

src/packages/core/src/metadata.ts
```diff
@@ -30,6 +32,10 @@ export interface EntityMetadata<G = unknown, D = unknown> {
 		// by fields. This is useful for entities that are not backed by a fully fledged provider or are
 		// otherwise dynamically generated.
 		excludeFromFiltering?: boolean;
+
+		// This means that there will be no recorded telemetry for this entity. This is useful for entities
```
thekevinbrown, line 36:
```suggestion
		// This means that there will be no recorded telemetry for this entity when it's requested from the Admin UI. This is useful for entities
```

src/packages/core/src/metadata.ts
```diff
@@ -30,6 +32,10 @@ export interface EntityMetadata<G = unknown, D = unknown> {
...
+		// that are managed by some other system or we don't want to record telemetry for.
```
thekevinbrown, line 37:
```suggestion
		// that are managed by some other system or we don't want to record telemetry for.
		//
		// Note that in order to achieve this from your own frontends, you need to set the `X-GRAPHWEAVER-SUPPRESS-TRACING: true` header on the requests, which is what this option instructs the admin UI to do for you.
```

src/packages/core/src/open-telemetry/entity.ts
```diff
@@ -0,0 +1,65 @@
...
+	);
+	Field(() => GraphQLNanoDuration, { adminUIOptions: { hideInFilterBar: true } })(
+		Trace.prototype,
+		'duration'
+	);
+	Field(() => String)(Trace.prototype, 'traceId');
+	Field(() => GraphQLJSON, { adminUIOptions: { hideInFilterBar: true, hideInTable: true } })(
+		Trace.prototype,
+		'attributes'
+	);
```
thekevinbrown, line 34: Could we just factory the whole entity here so our decorators look normal, or no?

thekevinbrown (approved):

src/packages/admin-ui-components/src/entity-list/styles.module.css
```diff
@@ -0,0 +1,73 @@
...
+	opacity: 0.6;
+}
+
+.successToast {
+	display: flex;
+	flex-direction: column;
+}
+
+.deletedText {
+	color: #ede8f2b2;
```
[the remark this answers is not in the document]
steve-keep: Not currently used anywhere else.

src/packages/core/src/open-telemetry/entity.ts
steve-keep: Yeah the reason for doing it this way is that the the developer can just apply ACL to the class as normal:

```
import { Trace } from '@exogee/graphweaver';
ApplyAccessControlList({
	DARK_SIDE: {
		// Dark side user role can perform operations on any tag
		all: true,
	},
})(Trace);
```

So it looks odd in this file but better from the end user.

PR #966, opened 2024-08-02
Fix / Ensure Error Logging in Lambda Mode

thekevinbrown:
Makes sure we log errors thrown when calling `startServerAndCreateLambdaHandler`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling and logging capabilities for the serverless function, improving reliability during startup.

- **Bug Fixes**
	- Ensured that initialization errors are properly logged and handled, preventing the function from proceeding with invalid configurations.


steve-keep (approved):

thekevinbrown:
I am not sure why this is breaking the tests. It works locally, but I don't want to have this hanging around unmerged, so we'll have another look later.

PR #1308, opened 2024-11-07
EXOG-386 Reusable custom hooks for entities

michaelleighstewart:
**Description:** Ability to add custom hooks at the entity level both in the `@Entity` decorator and with the `@Hook` decorator. 

**Example usage from `@Entity` decorator:**
```typescript
@Entity('User', {
	hooks: { [HookRegister.BEFORE_READ]: [hookFunction] },
	// ...
})
class User {}
```

**Example usage from `@Hook` decorator**
```typescript
@Hook(HookRegister.BEFORE_READ, hookFunction)
@Entity('User')
class User {}
```


thekevinbrown:
The only other thought I have on this is could we add the option to the `@Hook` decorator to attach an arbitrary function to an entity? I think currently it only can be applied when the function is a method, but we could dual-purpose it to also attach to an entity.

Issue #1190, opened 2024-09-23
Errors thrown from addUserToContext() are not logged clearly


thekevinbrown:
**Describe the bug**
From Slack: if `addUserToContext()` throws, we don't currently log that clearly, and instead throw a generic error.

**To Reproduce**
Throw from addUserToContext() and observe.

**Expected behavior**
Error can be generically returned from the API so we're not disclosing exactly what happened, but logs should be clear.

**Actual behavior**
Generic error is thrown, logs are not helpful.

**Additional context**
https://graphweaver.slack.com/archives/D0730MYM74Z/p1726103546839589?thread_ts=1723672001.215299&cid=D0730MYM74Z

thekevinbrown:
We believe this is resolved. If anyone is still experiencing poor logging, reach out, we're here to help!

PR #1330, opened 2024-11-12
Allow addUserToContext exceptions into logs

eddiesholl:
Small PR to address the issue recorded in https://github.com/exogee-technology/graphweaver/issues/1190. The message of an exception thrown anywhere during the auth processing in `authApolloPlugin` gets appended to the generic message `JWT verification failed. `.

There's many individual steps that can all individually fail within this `try` block. It could almost justify some [railway oriented programming](https://www.jonminter.dev/posts/railway-oriented-programming-async/) to handle the happy vs sad paths, that felt like a bit of a rabbit hole in the code base where it's not already used though.

### DIFF
--- a/src/packages/auth/src/authentication/apollo/plugin.ts
+++ b/src/packages/auth/src/authentication/apollo/plugin.ts
@@ -227,7 +227,7 @@ export const authApolloPlugin = <R>(
 
 						upsertAuthorizationContext(contextValue);
 					} catch (err: unknown) {
-						logger.trace(`JWT verification failed. ${err}`);
+						logger.error(`JWT verification failed. ${err}`);
 						tokenVerificationFailed = true;
 					}
 				}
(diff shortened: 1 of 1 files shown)


eddiesholl:
The solution right now is obviously the absolute smallest one to achieve the goal. It does mean the system is vulnerable to malicious activity hitting the logs, if requests were being made with invalid headers. We could mitigate that with more complex error handling with seperate try catch paths for trace logging vs error logging, for example.

src/packages/auth/src/authentication/apollo/plugin.ts
```diff
@@ -227,7 +227,7 @@ export const authApolloPlugin = <R>(
 
 						upsertAuthorizationContext(contextValue);
 					} catch (err: unknown) {
-						logger.trace(`JWT verification failed. ${err}`);
+						logger.error(`JWT verification failed. ${err}`);
```
thekevinbrown, line 230: Can we log the whole error instead of stringifying it, e.g.
```suggestion
						logger.error({ err }, `JWT verification failed.`);
```

thekevinbrown (changes requested):

src/packages/auth/src/authentication/apollo/plugin.ts
eddiesholl: Yeah good call, I've made that change locally to check the output and pushed it up

thekevinbrown (approved):

PR #1471, opened 2024-12-23
Fix / Correctly differentiate between creates and updates with numeric primary key

thekevinbrown:
When the DB would return a number for the primary key, on updates we'd incorrectly see these updates as creates because we couldn't locate the updated entities due to a number not being === to a string.

This always casts primary keys to strings so they will be the same type. In addition this moves us to a `Set` for this comparison so we're O(1) for the compares instead of O(n).

### DIFF
--- a/src/packages/auth/src/auth-utils.ts
+++ b/src/packages/auth/src/auth-utils.ts
@@ -251,13 +251,19 @@ export async function checkAuthorization<G = unknown>(
 	requestInput: Partial<G>,
 	requiredPermission: AccessType
 ) {
+	logger.trace({ entityName, id, requestInput, requiredPermission }, 'Entering checkAuthorization');
+
 	// Get ACL first
 	const acl = getACL(entityName);
 	const meta = graphweaverMetadata.getEntityByName(entityName);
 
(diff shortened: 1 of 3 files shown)


src/packages/auth/src/auth-utils.ts
```diff
@@ -251,14 +251,20 @@ export async function checkAuthorization<G = unknown>(
 	requestInput: Partial<G>,
 	requiredPermission: AccessType
 ) {
+	logger.trace({ entityName, id, requestInput, requiredPermission }, 'Entering checkAuthorization');
+
 	// Get ACL first
 	const acl = getACL(entityName);
 	const meta = graphweaverMetadata.getEntityByName(entityName);
 
+	logger.trace('Checking whether user can perform reuqested action.');
```
eddiesholl, line 260: Typo

eddiesholl (approved):
Nice!

src/packages/auth/src/auth-utils.ts
thekevinbrown: Fixed!

juan871003 (approved):

PR #1526, opened 2025-01-17
Implement new simple text filter

eddiesholl:
Adding a new filter type for the admin-ui-components filter bar. This is a simple text input, it has no auto complete, it just fetches with the value you have typed in. You can get an exact match if your filter value is correct. This is to help with usability issues on large list views.

There is a relatively invasive UI change here. The entity-list currently replaces the content area with a big loading image while a new filter is being applied to the query. This mens that your typing location and focus is being thrown away. It's fairly unnecessary, as the table area has its own spinner, so I'm just removing the use of the big loader image.

There is technically a breaking change in here, as the meaning of AdminUIFilterType.TEXT has changed with no migration path

I've noticed that there are duplicate enum definitions for AdminUIFilterType in the core and admin-ui-components packages. There's no dependency between these packages either way right now. I assume the solution will be to create a shared package for this type. I have no idea how many other types might be in this duplicated state


src/packages/admin-ui-components/src/filters/dropdown-text-filter.tsx
```diff
@@ -0,0 +1,57 @@
+import { useLazyQuery } from '@apollo/client';
```
eddiesholl, line 1: Straight clone of the previous TextFilter file

src/packages/admin-ui-components/src/input/component.tsx
```diff
@@ -30,6 +30,11 @@ export const Input = ({
 	const [value, setValue] = useState<string | undefined>(initialValue ?? '');
 	const debouncedValue = useDebounce<string | undefined>(value, 800); // debounce request for 800ms
 
+	useEffect(() => {
```
eddiesholl, line 33: Without this change, the new TextFilter does not empty its value when clicking on 'Empty FIlters'. Another solution would be we have an uncontrolled Input component that does not track state like this

thekevinbrown: I think it should still be a controlled input, just not have its own state, e.g. use the filter to hold the state the way the rest of the filter components do.

src/packages/core/src/metadata-service/resolver.ts
```diff
@@ -18,17 +18,17 @@ const mapFilterType = (field: AdminUiFieldMetadata): AdminUIFilterType => {
 	// Otherwise check the type
 	switch (field.type) {
 		case 'ID':
-			return AdminUIFilterType.TEXT;
+			return AdminUIFilterType.DROP_DOWN_TEXT;
 		case 'Number':
 			return AdminUIFilterType.NUMERIC;
 		case 'String':
-			return AdminUIFilterType.TEXT;
+			return AdminUIFilterType.DROP_DOWN_TEXT;
```
thekevinbrown, line 25: The default should be `TEXT` for both `ID` and `String` I reckon.

thekevinbrown (changes requested):
Two things:
- We should make the filter component stateless the way it is for other filters, e.g. it just stores its state in the filter itself.
- We should make the default `TEXT` not `DROP_DOWN_TEXT` for `ID` and `String` types.

src/examples/rest/src/backend/schema/person.ts
```diff
@@ -17,13 +17,16 @@ import { Vehicle } from './vehicle';
...
 export class Person {
-	@Field(() => ID, { primaryKeyField: true })
+	@Field(() => ID, {
+		primaryKeyField: true,
+		adminUIOptions: { filterType: AdminUIFilterType.DROP_DOWN_TEXT },
+	})
 	url!: string;
 
-	@Field(() => String)
+	@Field(() => String, { adminUIOptions: { filterType: AdminUIFilterType.TEXT } })
```
thekevinbrown, line 26: Since this is the default let's just do:
```suggestion
	@Field(() => String)
```

thekevinbrown (approved):

Issue #1563, opened 2025-02-03
Make filter placeholders match field names


mackcheesman:
Describe the bug
Some filter components (e.g. ComboBox, DatePicker) capitalise the field name in the placeholder text/label.

To Reproduce
Run the sqlite example project in this repo, and view the Artist table. The ComboBox to filter by artist has the placeholder text 'ArtistId'

Expected behavior
By default, it should match the GraphQL entity field name.

Actual behavior
Field names are being capitalised

PR #1564, opened 2025-02-03
Removed capitalisation transformation CSS from filter components

mackcheesman:
https://github.com/exogee-technology/graphweaver/issues/1563

Removed some CSS that was capitalising the placeholders for some input components used for filtering on the Admin UI.

Now the field names should match the GraphQL entity field names by default.


thekevinbrown (commented):
Thanks!

thekevinbrown (approved):

PR #1565, opened 2025-02-03
Removed capitalisation in Admin UI filter inputs

mackcheesman:
Re-opening this PR on the main repo instead of a fork while the issue with auth test secrets remains unresolved.

https://github.com/exogee-technology/graphweaver/issues/1563

Removed some CSS that was capitalising the placeholders for some input components used for filtering on the Admin UI.

Now the field names should match the GraphQL entity field names by default.


thekevinbrown:
Thanks!

PR #1570, opened 2025-02-04
Add Vitest configuration and logging enhancements for sensitive data

juan871003:
Added a custom callback function to `graphweaverMetadata.addMutation` to log any incoming queries taking care of possible personal data that should be obfuscated.


src/packages/auth/src/authentication/methods/utils.ts
```diff
@@ -67,3 +69,47 @@ export const runAfterHooks = async <D, H = CredentialInsertInput | CredentialUpd
 
 	return hookEntities;
 };
+
+export const makeLoginPasswordQuerySafeForLogging = (
+	queryString: string,
+	variables: VariableValues | undefined
+) => {
+	const safeVariables: VariableValues = JSON.parse(JSON.stringify(variables ?? {}));
+	const ast = parse(queryString);
```
thekevinbrown, line 78: I think we should already have this value passed to us parsed if we're in `didResolveOperation`, no?

src/packages/auth/src/authentication/methods/utils.ts
```diff
@@ -67,3 +69,47 @@ export const runAfterHooks = async <D, H = CredentialInsertInput | CredentialUpd
...
+						value: '********',
+					},
+				};
+			}
+
+			if (
+				node.kind === Kind.ARGUMENT &&
+				node.name.value === 'password' &&
+				node.value.kind === Kind.VARIABLE
+			) {
```
thekevinbrown, line 100: Would be a little bit more readable I think to have one big `if`  with `node.kind === Kind.ARGUMENT && node.name.value === 'password'` then inside have the check for `node.value.kind`.

src/packages/auth/src/authentication/methods/utils.ts
```diff
@@ -67,3 +69,47 @@ export const runAfterHooks = async <D, H = CredentialInsertInput | CredentialUpd
 
 	return hookEntities;
 };
+
+export const makeLoginPasswordQuerySafeForLogging = (
```
thekevinbrown, line 73: What about a naming it `maskSensitiveValuesForLogging`?

thekevinbrown (changes requested):
The only thing I'm concerned about here is the potential double parsing of the GraphQL. Apollo should be doing the parsing for us and should be passing us an AST as far as I'm aware.

src/packages/auth/src/authentication/methods/utils.ts
juan871003: Totally, I don't know how that didn't occur to me.

thekevinbrown (approved):

Issue #1458, opened 2024-12-18
Nested entity from a custom mutation result fails at runtime


eddiesholl:
**Describe the bug**
The resolver for a nested entity (batchCode in this case) is failing at runtime when I have selected a list of entities with a child entity. The selection is coming off the result of a custom mutation.

 ```
Cannot read properties of undefined (reading 'batchCode')\n    at id .../src/packages/backend/.graphweaver/backend/index.js:768:51)\n    at _listRelationshipField (.../src/node_modules/.pnpm/@exogee+graphweaver@2.9.7_encoding@0.1.13/node_modules/@exogee/graphweaver/lib/resolvers.js:380:15)```

I believe the problem is at this line of code. The source must be a graphql entity that does not have the 'dataEntity' populated
https://github.com/exogee-technology/graphweaver/blob/c5cfe2a726c71a1f8435f93ea5ce269d5c404faf/src/packages/core/src/resolvers.ts#L566

**To Reproduce**
Steps to reproduce the behavior:

1. Define an entity that has a nested entity
2. Create a custom mutation that returns a list of the parent entities
3. The mutation should return microform entity instances
4. Select the 2 levels of entities

**Expected behavior**
This should not fail at runtime

**Actual behavior**
It throws an error

**Screenshots or Screencasts**


**Desktop (please complete the following information):**

- MacOS 14.4

**Additional context**
@thekevinbrown has a pretty good sense of what is probably wrong here

PR #1623, opened 2025-02-25
Exogw-400 - add a test and example for a custom mutation with a nested result

eddiesholl:
This PR is filling in some gaps for users trying to understand how to add custom queries and mutations. I've added an example of both to the existing storage provider example, and then exercised those new items with e2e tests.

I've tried to keep the examples realistic and detailed enough to be actually useful.

In parallel with this work, I've also added more detail to the docs page for custom items at https://graphweaver.com/docs/customising-base-resolver

Addressing https://github.com/exogee-technology/graphweaver/issues/1458


src/packages/storage-provider/src/decorators/media-field.ts
```diff
@@ -66,10 +66,10 @@ export class GraphweaverMedia {
 	static serialize = ({ value }: { value: unknown }) => {
 		if (value === null) return null;
 		if (isMedia(value)) {
-			return JSON.stringify({
+			return {
```
eddiesholl, line 69: I'm pretty certain this change is needed here. The Postgres driver is accepting the object, and it gets persisted in a format that allows key based querying from the jsonb column.

There is a potential matching change in the deserialise below, but I have left it there for backwards compatibility. It tries a json parse of the untyped incoming value is detected as a string.

eddiesholl: I suppose this is a breaking change, as existing users would now see differently formatted results in their underlying JSON and JSONB columns, although the deserialise will transparently handle either.

But an independent service reading these columns could potentially be broken by the change

src/examples/s3-storage/src/backend/custom-mutations/index.ts
```diff
@@ -0,0 +1,86 @@
...
+		const upload = await s3.getUploadUrl({
+			args: { key: filename },
+			source,
+			context,
+			info,
+			fields,
+		});
+
+		// upload the image to s3
+		await fetch(upload.url, {
```
thekevinbrown, line 67: I think we'll need to get a signed URL here for the upload or we'll get a 403 on AWS. In Minio this probably works, but I doubt it would in prod.

src/packages/end-to-end/src/__tests__/api/storage-provider/issues/gh1458.test.ts
```diff
@@ -0,0 +1,144 @@
...
+		}
+		await fetch(uploadUrl, {
+			method: 'PUT',
+			body: file,
+		});
+
+		const newSubmission = await request<{ createSubmission: Submission }>(config.baseUrl).path('/')
+			.query(gql`
+			mutation {
+				createSubmission(input: { image: { filename: "${uploadedFilename}", type: IMAGE } }) {
```
thekevinbrown, line 50: I think you can pass this in as a variable to the `.query()` call, no? Would be a bit more maintainable.

src/packages/end-to-end/src/__tests__/api/storage-provider/issues/gh1458.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+		testSubmission = newSubmission.data?.createSubmission;
+	});
+
+	describe('Custom queries', () => {
+		test('should allow a selection of nested entities in a custom query', async () => {
+			const response = await request<{ submissionByFilename: Submission }>(config.baseUrl).path('/')
+				.query(gql`
+				query {
+					submissionByFilename(filename: "${testSubmission.image.filename}") {
```
thekevinbrown, line 72: Same here.

src/packages/end-to-end/src/__tests__/api/storage-provider/issues/gh1458.test.ts
```diff
@@ -0,0 +1,144 @@
...
+	});
+
+	describe('Custom mutations', () => {
+		test('should allow a selection of nested entities in custom mutations', async () => {
+			const newSubmissionId = testSubmission.id;
+
+			const response = await request<{ createThumbnail: Submission }>(config.baseUrl).path('/')
+				.query(gql`
+				mutation {
+					createThumbnail(input: { submissionId: "${newSubmissionId}", width: 100, height: 100 }) {
```
thekevinbrown, line 117: Same here.

src/packages/storage-provider/src/decorators/media-field.ts
thekevinbrown: If we were persisting a string before instead of a JSON object in the JSONB column then that was 100% a bug and we should fix it.

The only other thought I have is we should give them a script they can run to update all the JSONB strings to actual JSONB values in place so they don't have to mutate them all through the backend to get them over to the originally intended format.

Also sounds like we should just remove the serialize / de-serialize from this entity functionality altogether eventually.

I'll note this in the release notes, thanks for bringing it to my attention!

thekevinbrown (commented):
Only things that worry me are the variable passing which should be pretty quick to fix up, looks good!

src/examples/s3-storage/src/backend/custom-mutations/index.ts
eddiesholl: Thankfully whoever implemented getUploadUrl originally in the s3 provider already thought about that 😅

src/packages/storage-provider/src/decorators/media-field.ts
eddiesholl: I've had a think about all the different ways we could 'script' that migration. It seems unlikely that a user would want to run the graphweaver CLI against their production app. The best I can come up with is to offer a new action button in the admin UI 'migrate data'. This could pop open a modal, and you could choose:
- column to migrate
- source data type
- target data type

What do you think? I can't think of a more 'kitchen sink included' style solution. I think anyone with real production data is probably running proper schema migration steps anyway, which is probably a better format for the fix.

image

thekevinbrown: I just mean an example SQL script that we can pop in the release notes, sorry, should have been more clear!

src/examples/s3-storage/src/backend/custom-mutations/index.ts
thekevinbrown: Nice.

src/packages/storage-provider/src/decorators/media-field.ts
eddiesholl: OK I think this is what we're after, I think the PR is good to go now:
```
UPDATE submission
SET image = 
  CASE 
    WHEN jsonb_typeof(image) = 'string' THEN (image#>>'{}')::jsonb
    ELSE image
  END
WHERE jsonb_typeof(image) = 'string';
```

PR #1624, opened 2025-02-26
Obfuscate sensitive fields in authentication mutations

juan871003:
Implement a mechanism to obfuscate sensitive argument values in authentication-related mutations to enhance security during logging. This change applies to various mutations across different authentication methods.

### DIFF
--- a/src/packages/auth/src/authentication/methods/utils.ts
+++ b/src/packages/auth/src/authentication/methods/utils.ts
@@ -89,30 +89,32 @@ const obfuscateSensitiveValues = (
 				return '********';
 			}
 
-			if (value === null || value === undefined) return value;
+			if (value === null || value === undefined) {
+				return value;
+			}
 
 			if (Array.isArray(value)) {
 				return value.map((item: any) => obfuscateNode(key, item));
 			}
 
-			if (typeof value === 'symbol') return '********'; // technically impossible, but just in case
+			if (typeof value === 'symbol') {
+				return '********'; // technically impossible, but just in case
+			}
 
 			if (typeof value === 'object') {
-				for (const newKey in value) {
-					value[newKey] = obfuscateNode(newKey, value[newKey]);
-				}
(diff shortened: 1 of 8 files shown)


src/packages/auth/src/authentication/methods/utils.ts
```diff
@@ -70,14 +77,63 @@ export const runAfterHooks = async <D, H = CredentialInsertInput | CredentialUpd
...
+			if (typeof value === 'object') {
+				for (const newKey in value) {
+					value[newKey] = obfuscateNode(newKey, value[newKey]);
+				}
+			}
+
+			return value;
+		};
+
+		const variablesCopy = JSON.parse(JSON.stringify(variables));
```
eddiesholl, line 109: It's cleaner to clone the object instance with { ...variables }

src/packages/auth/src/authentication/methods/utils.ts
```diff
@@ -70,14 +77,63 @@ export const runAfterHooks = async <D, H = CredentialInsertInput | CredentialUpd
...
+
+		for (const key in variablesCopy) {
```
eddiesholl, line 111: I prefer to combine the clone and then value iteration using Object.fromEntries, like in this example:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#object_transformations

src/packages/auth/src/authentication/methods/utils.ts
thekevinbrown: Yeah, but that does a shallow copy. There are nested values then the references will still be attached. @juan871003 are there ever nested values in the variables object?

thekevinbrown (commented):
I'm happy with this, but agree with Eddie, if a shallow copy is all that's needed then a shallow copy would be better.

thekevinbrown (approved):

PR #1625, opened 2025-02-26
Add handling for field_null and field_notnull in gqlToMikro

juan871003:
Introduce support for null and notnull filters in the gqlToMikro function, allowing for more flexible querying of data based on nullability. Add corresponding tests to validate the new functionality.

### DIFF
--- a/src/packages/end-to-end/src/__tests__/api/postgres/query/null-filter.test.ts
+++ b/src/packages/end-to-end/src/__tests__/api/postgres/query/null-filter.test.ts
@@ -1,0 +1,26 @@
+import request from 'supertest-graphql';
+import gql from 'graphql-tag';
+
+import { config } from '../../../../config';
+
+describe('null filter', () => {
+	test('should filter by Customers with company_null = true', async () => {
+		const { data } = await request<{ customers: any }>(config.baseUrl)
+			.query(gql`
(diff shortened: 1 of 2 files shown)


juan871003 (commented):
https://exogee.atlassian.net/browse/EXOGW-403

src/packages/end-to-end/src/__tests__/api/postgres/query/null-filter.test.ts
```diff
@@ -0,0 +1,26 @@
...
+				query Customers($filter: CustomersListFilter) {
+					customers(filter: $filter) {
+						customerId
+						company
+					}
+				}
+			`)
+			.variables({
+				filter: {
+					company_null: true,
```
thekevinbrown, line 19: Can we also add a similar test for `company_notnull: true` with an expected list of values?

thekevinbrown (changes requested):

src/packages/end-to-end/src/__tests__/api/postgres/query/null-filter.test.ts
juan871003: Yep, I'm just checking if the test passes first. Then I'll add more. But I don't see this test being run at all.

thekevinbrown (approved):

PR #1631, opened 2025-02-27
Exogw-395 - Implement bigint columns as string during import

eddiesholl:
When we import a database with a `bigint` or `bigserial` type column, the project is not left in a state where you can query from that entity type. By default, we treat a `bigint` as a numerical type, but then when it comes time for graphql to try to represent the `bigint` on the wire, it fails to be represented as an ID type.

This is adding very special case handling for `bigint`, so it gets represented as a string.

There's another path you can take, which is to use a scalar like https://the-guild.dev/graphql/scalars/docs/scalars/big-int, that would introduce a larger extra dependency though.

I've included this solution in the 'databases' example for exposing a `bigint` as an actual number.


eddiesholl:
There isn't any mention of scalars in the current docs. Do you think a new page is needed, and it can cover this complexity?

eddiesholl:
Also this PR is taking the approach of exposing "all the solutions", not sure if that's a smell and our import should do the same thing as the example, and use a custom scalar instead

src/examples/databases/src/backend/entities/mysql/task.ts
```diff
@@ -1,10 +1,10 @@
-import { BigIntType, Entity, PrimaryKey, Property } from '@mikro-orm/core';
 import { ExternalIdField } from '@exogee/graphweaver-mikroorm';
+import { BigIntType, Entity, PrimaryKey, Property } from '@mikro-orm/core';
 
 @Entity()
 export class Task {
-	@PrimaryKey({ type: new BigIntType('string') })
-	id!: string;
+	@PrimaryKey({ type: new BigIntType() })
+	id!: number;
```
thekevinbrown, line 7:
```suggestion
	id!: bigint;
```
I don't think `number` can represent the full range of `bigint` values.

src/examples/databases/src/backend/schema/task.ts
```diff
@@ -1,23 +1,22 @@
...
-import { myConnection } from '../database';
 
 @Entity('Task', {
 	provider: new MikroBackendProvider(OrmTask, myConnection),
 })
 export class Task {
-	@Field(() => ID)
-	id!: string;
+	@Field(() => GraphQLBigInt)
+	id!: number;
```
thekevinbrown, line 19:
```suggestion
	id!: bigint;
```

src/examples/databases/src/frontend/types.generated.ts
```diff
@@ -385,14 +387,10 @@ export type TasksListFilter = {
...
-  id_gte?: InputMaybe<Scalars['ID']['input']>;
-  id_in?: InputMaybe<Array<Scalars['ID']['input']>>;
-  id_lt?: InputMaybe<Scalars['ID']['input']>;
-  id_lte?: InputMaybe<Scalars['ID']['input']>;
-  id_ne?: InputMaybe<Scalars['ID']['input']>;
-  id_nin?: InputMaybe<Array<Scalars['ID']['input']>>;
+  id?: InputMaybe<Scalars['BigInt']['input']>;
+  id_in?: InputMaybe<Array<Scalars['BigInt']['input']>>;
+  id_ne?: InputMaybe<Scalars['BigInt']['input']>;
+  id_nin?: InputMaybe<Array<Scalars['BigInt']['input']>>;
```
thekevinbrown, line 393: I reckon we should still be generating the comparison operators for `BigInt` scalars. Maybe that's also special handling in the Schema Generator?

src/packages/mikroorm/src/introspection/files/data-entity-file.ts
```diff
@@ -128,6 +128,10 @@ export class DataEntityFile extends BaseFile {
 			}
 		}
 
+		if (prop.type === 'bigint') {
```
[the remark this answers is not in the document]
thekevinbrown: Yeah most of the generation code is just "When we see X do thing" and I think abstracting it will just make it harder to read. This is a good way to do things.

thekevinbrown (approved):
I think the only thing that'd be great to do is implement the comparison operators for BigInt scalars specifically on filters, but that's a different task.

src/examples/databases/src/frontend/types.generated.ts
thekevinbrown: Decided this is a different task, all good!

thekevinbrown (approved):

PR #1632, opened 2025-02-27
`EXOGW-415` - gqlToMikro -> handle multiple operations on the same filter key

juan871003:
https://exogee.atlassian.net/browse/EXOGW-415

Before this change we were getting

```
{
  "0": "t",
  "1": "e",
  "2": "s",
  "3": "t",
  $nin: [
    "not here",
  ],
}
```

when sending a filter like:
```
users(filter:  {
     notes_nin: ["not here"],
     notes: "test"
  }) {
    id
  }
```

Now we handle this situation and some others as well.
Unit tests were also added highlighting some of the scenarios now handled properly

### DIFF
--- a/src/packages/mikroorm/src/provider/provider.test.ts
+++ b/src/packages/mikroorm/src/provider/provider.test.ts
@@ -1,23 +1,21 @@
 import { describe, it, expect } from 'vitest';
-
 import { gqlToMikro } from './provider';
 
-const filters = [
-	{
-		key: 'nested -> with numbers',
-		throws: false,
-		test: {
+describe('gqlToMikro', () => {
+	it('transforms filter: nested -> with numbers', () => {
+		const test = {
 			field1_gt: 1,
 			field1_lt: 2,
 			field1_in: [1, 2, 3],
 			nested: {
 				field1: 1,
+
 				field2_gte: 3,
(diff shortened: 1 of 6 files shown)


src/packages/mikroorm/src/provider/provider.test.ts
```diff
@@ -0,0 +1,212 @@
...
+					subsubfield1: 'string2',
+					subsubfield2: { $gt: 7 },
+				},
+			},
+		},
+	},
+];
+
+describe('gqlToMikro', () => {
+	filters.forEach(({ key, test, expectedResult, throws }) => {
```
thekevinbrown, line 203: This is exactly the opposite feedback that you'd normally get, but in tests specifically I'm a huge fan of copy/paste instead of trying to create abstractions like this. Can you copy / paste the actual values into separate tests? Will make it easier to maintain them should we need to refactor as they won't have to stick to this format if it doesn't make sense for them to.

thekevinbrown (changes requested):
This is looking great, let's just make the tests a bit simpler and we're ready to ship once the tests pass.

juan871003:
https://exogee.atlassian.net/browse/EXOGW-415

thekevinbrown (approved):

Issue #1583, opened 2025-02-10
*_ilike filter operator doesn't work with SQLite Backend Provider


mackcheesman:
**Describe the bug**
SQLite doesn't have `ILIKE`; it has `LIKE` which is case insensitive by default ([but can be configured to be case sensitive](https://www.sqlite.org/pragma.html)).

When using the generated filters for an entity, the `_like` operator works and is case insensitive unless the connected SQLite DB has been configured otherwise. Using the `_ilike` operator will cause an error:
```json
{ "message": "near \"ilike\": syntax error" }
```

**To Reproduce**

1. Run the SQLite example in this repo (`src/examples/sqlite`)
2. Open the Admin UI
3. Go to the playground
4. Run this query:
```
query AlbumsList($filter: AlbumsListFilter) {
  albums(filter: $filter) {
    albumId
    title
  }
}
```
with these variables:
```json
{
  "filter": {
    "title_ilike": "% AND %"
  }
}
```
**Actual behavior**
The query fails with this error:
```json
{
  "errors": [
    {
      "message": "near \"ilike\": syntax error",
      "locations": [
        {
          "line": 33,
          "column": 3
        }
      ],
      "path": [
        "albums"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ],
}
```

**Expected behavior**
The query should at least succeed. There's a number of reasonable ways this could be handled, including (but not limited to):
- Not having `_ilike` on the filters for any entities from the SQLite backend provider
 - Only have `_like` (or `_ilike`, and have it use `LIKE` under the hood)
- Implement support for both `_like` and `_ilike` for the SQLite backend provider

**Desktop (please complete the following information):**
- SQLite version: 3.45.0 (`3.45.0 2024-01-15 17:01:13 1066602b2b1976fe58b5150777cced894af17c803e068f5918390d6915b46e1d (64-bit)`)
- OS and version: MacOS 15.3

thekevinbrown:
Proposal: When connecting to an SQLite database issue `PRAGMA case_sensitive_like=ON;` on connection.

Later, when querying, leave the `like` transform alone, but `ilike` should translate to `UPPER(field) LIKE (UPPER(pattern)`.

PR #1641, opened 2025-03-03
Fix (Mikroorm package): Handling ILIKE operator for non-postgres databases

mackcheesman:
https://github.com/exogee-technology/graphweaver/issues/1583

ILIKE isn't supported for MySQL or SQLite. Since LIKE is case-insensitive (mostly). This PR introduces extra logic around filtering so that `column_ilike` filters are instead coerced to `column_like`.


src/packages/mikroorm/src/provider/provider.ts
```diff
@@ -165,6 +171,24 @@ export class MikroBackendProvider<D> implements BackendProvider<D> {
...
+				return "mysql";
+			case PostgreSqlDriver.name:
+				return "postgresql";
+			default:
+				throw new Error(`This driver (${driver}) is not supported!`);
+		}
+	}
+
+	private gqlToMikro(filter: any) {
+		return gqlToMikro(filter, this.getDbType());
```
thekevinbrown, line 191: I'd prefer to get rid of this and just have wherever we call `gqlToMikro` look like this. This is just to avoid the confusion between `gqlToMikro` and `this.gqlToMikro`.

src/packages/mikroorm/src/provider/provider.ts
```diff
@@ -323,7 +347,7 @@ export class MikroBackendProvider<D> implements BackendProvider<D> {
 		// This query only works if we JSON.parse(JSON.stringify(filter)):
 		const where = traceSync((trace?: TraceOptions) => {
 			trace?.span.updateName('Convert filter to Mikro-Orm format');
-			return filter ? gqlToMikro(JSON.parse(JSON.stringify(filter))) : undefined;
+			return filter ? this.gqlToMikro(JSON.parse(JSON.stringify(filter))) : undefined;
```
thekevinbrown, line 350: So in other words:

```suggestion
			return filter ? gqlToMikro(JSON.parse(JSON.stringify(filter)), this.getDbType()) : undefined;
```

thekevinbrown (changes requested):
Looking great! One small code structure thing to fix up and then we're good to go!

thekevinbrown (approved):

PR #1655, opened 2025-03-05
`EXOGW-419` and `EXOGW-416`

juan871003:
https://exogee.atlassian.net/browse/EXOGW-416
and
https://exogee.atlassian.net/browse/EXOGW-419

- deep copy `filter` in mikroorm provider because that filter is not a POJO and mikroorm doesn't generate the proper SQL query
- use the filter coming from BEFORE_READ hook in the `core` resolver in the `_listRelationshipField` and `aggregateRelationshipField` so we apply ACLs to the resulting SQL query

### DIFF
--- a/src/packages/core/src/metadata-service/resolver.utils.ts
+++ b/src/packages/core/src/metadata-service/resolver.utils.ts
@@ -1,0 +1,220 @@
+import { BaseContext } from '@apollo/server';
+import { dataEntityForGraphQLEntity } from '../default-from-backend-entity';
+import { Filter, ReadHookParams, ResolverOptions, TraceOptions } from '../types';
+import { EntityMetadata, MetadataType } from '../metadata';
+import { isScalarType } from 'graphql';
+import { getFieldType } from '../schema-builder';
+import { logger } from '@exogee/logger';
+import { BaseLoaders } from '../base-loader';
+import { isDefined } from '../utils';
+
+export type ID = string | number | bigint;
+
+/**
+ * Retrieves the ID value from the source data.
+ * If the id is a function then it calls it to get the ID value.
+ * If the id is a string then it tries to get the value from the source data.
+ */
+export const getIdValue = <G, D, R, C extends BaseContext>(
+	id: string | bigint | ((dataEntity: unknown) => string | number | bigint | undefined) | undefined,
+	source: G
+) => {
+	let idValue: ID | ID[] | undefined = undefined;
+	if (id && typeof id === 'function') {
+		// If the id is a function, we'll call it with the source data to get the id value.
+		idValue = id(dataEntityForGraphQLEntity<G, D>(source as any));
+	} else if (id) {
+		// else if the id is a string, we'll try to get the value from the source data.
+		const valueOfForeignKey = dataEntityForGraphQLEntity<G, D>(source as any)?.[id as keyof D];
+
+		// If the value is a string or number, we'll use it as the id value.
+		if (
+			typeof valueOfForeignKey === 'string' ||
+			typeof valueOfForeignKey === 'number' ||
+			typeof valueOfForeignKey === 'bigint' ||
+			Array.isArray(valueOfForeignKey)
+		) {
+			idValue = valueOfForeignKey;
+		} else if (!isDefined(valueOfForeignKey)) {
+			// If the value is null, we'll use it as the id value.
+			idValue = undefined;
+		} else {
+			// The ID value must be a string or a number otherwise we'll throw an error.
+			throw new Error(
+				'Could not determine ID value for relationship field. Only strings, numbers or arrays of strings or numbers are supported.'
+			);
+		}
+	}
+
+	return idValue;
+};
+
+interface ConstructFilterForRelatedEntityParams<G, D, R, C extends BaseContext> {
+	resolverOptions: ResolverOptions<{ filter: Filter<R> }, C, G>;
+	idValue: ID | ID[] | undefined;
+	relatedPrimaryKeyField: string;
+	relatedField: string | undefined;
+	relatedEntityMetadata: EntityMetadata;
+	sourcePrimaryKeyField: keyof G;
+}
+
+/**
+ * For this entity we may get a filter, this is the filter passed by the client for this entity,
+ * we then get this filter and _and it with the filter for the relationship.
+ * Both the resulting filter and the relationship filter are returned so that later we can remove the relationship filter from the filter.
+ */
+export const constructFilterForRelatedEntity = <G, D, R, C extends BaseContext>(
+	params: ConstructFilterForRelatedEntityParams<G, D, R, C>
+) => {
+	const {
+		resolverOptions: {
+			source,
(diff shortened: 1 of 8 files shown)


src/packages/core/src/resolvers.ts
```diff
@@ -686,6 +692,44 @@ const _listRelationshipField = async <G, D, R, C extends BaseContext>(
 
 	logger.trace('Loading from BaseLoaders');
 
+	let relationshipFilterChunkFound = !relationshipFilterChunk;
+	/**
+	 * Look for `relationshipFilterChunk` and remove it from the filter.
+	 * This is necessary because we are batch loading data and that relationshipFilterChunk makes the filter different for each record, which ends up triggering a SQL query for each record (not batching at all).
+	 * What we want to achieve is having a filter with ACLs into it, but without the relationship chunk.
+	 */
+	const removeRelationshipFilterChunk = (
```
eddiesholl, line 701: Is there any reason to define this as an inline function instance? Could you just make it a statement that is calculated and assigned to `loadFilter`?

src/packages/mikroorm/src/provider/provider.ts
```diff
@@ -462,10 +462,10 @@ export class MikroBackendProvider<D> implements BackendProvider<D> {
 		let queryFilter: any = { [relatedField]: { $in: relatedFieldIds } };
 
 		if (filter) {
+			// JSON.parse(JSON.stringify()) is needed. See https://exogee.atlassian.net/browse/EXOGW-419
+			const gqlToMikroFilter = JSON.parse(JSON.stringify([gqlToMikro(filter, this.getDbType())]));
```
eddiesholl, line 466: Just curious what sort of keys and values are you trying to remove here?

src/packages/core/src/resolvers.ts
juan871003: It's a recursive function

src/packages/mikroorm/src/provider/provider.ts
juan871003: We are not trying to remove any properties from the queryFilter object. However, Mikro-ORM internally checks if the object is a POJO or if it was constructed from a class or has any other inherited methods that are not just plain properties. Mikro-ORM will treat filters as plain objects and non-plain objects differently, and the resulting SQL Query was wrong.
The `filter` object coming as a param to this `findByRelatedId` method is not a plain object, it comes directly from GraphQL land, so the resulting SQL was not as we expected. Parse-stringifying this object will give as a POJO which resulting in Mikro-ORM giving us the correct SQL.

src/packages/core/src/metadata-service/resolver.utils.ts
```diff
@@ -0,0 +1,218 @@
...
+	 * This is necessary because we are batch loading data and that relationshipFilterChunk makes the filter different for each record, which ends up triggering a SQL query for each record (not batching at all).
+	 * What we want to achieve is having a filter with ACLs into it, but without the relationship chunk.
+	 */
+	const removeRelationshipFilterChunk = (
+		filter: Filter<R> | undefined,
+		relationshipFilterChunk: Filter<R>
+	): Filter<R> | undefined => {
+		if (!filter?._and) return filter;
+		return {
+			_and: filter._and.map((item) => {
```
thekevinbrown, line 146: Probably worth adding a comment here that the reason we're only looking through `_and` is because the core code puts the join in an `_and` filter. Other filters are supported and will work fine with this code, we just know that the relationship filter won't be in there.

src/packages/end-to-end/src/__tests__/api/auth/acl/EXOGW-416.test.ts
```diff
@@ -0,0 +1,274 @@
...
+let token: string | undefined;
+let em: EntityManager | undefined = undefined;
+
+describe('Nested entity queries should not bypass row-level security', () => {
+	beforeAll(async () => {
+		const connectionResult = await ConnectionManager.connect('sqlite', connection);
+		em = connectionResult?.em;
+		assert(em !== undefined);
+
+		// If tables exists then fail, we are assuming that the tables do not exist
```
thekevinbrown, line 141:
```suggestion
		// If tables exist then fail, we are assuming that the tables do not exist
```

src/packages/end-to-end/src/__tests__/api/auth/acl/EXOGW-416.test.ts
```diff
@@ -0,0 +1,274 @@
...
+			.execute('SELECT name FROM sqlite_master WHERE type="table"');
+		const tableNames = tables.map((table: any) => table.name);
+
+		if (
+			tableNames.includes('Task') ||
+			tableNames.includes('Tag') ||
+			tableNames.includes('task_tags')
+		) {
+			throw new Error(
+				'Tables already exist, this test suit assumes that the tables do not exist, the suit throws away the tables at the end of the test'
```
thekevinbrown, line 153:
```suggestion
				'Tables already exist, this test suite assumes that the tables do not exist. The suite throws away the tables at the end of the test'
```

src/packages/end-to-end/src/__tests__/api/auth/acl/EXOGW-416.test.ts
```diff
@@ -0,0 +1,274 @@
...
+		const tagForAuthenticatedUser = new OrmTag('tagForAuthenticatedUser');
+		tagForAuthenticatedUser.tasks.add(task1ForAuthenticatedUser);
+		tagForAuthenticatedUser.tasks.add(task2ForAuthenticatedUser);
+		tagForAuthenticatedUser.tasks.add(task3ForAuthenticatedUser);
+		const tagForOtherUser = new OrmTag('tagForOtherUser');
+		tagForOtherUser.tasks.add(task1ForOtherUser);
+		tagForOtherUser.tasks.add(task2ForOtherUser);
+		tagForOtherUser.tasks.add(task3ForOtherUser);
+
+		await em?.persistAndFlush([
```
thekevinbrown, line 225:
```suggestion
		await em!.persistAndFlush([
```

Probably makes sense for this to throw if it can't run because `em` is undefined.

src/packages/end-to-end/src/__tests__/api/auth/acl/EXOGW-416.test.ts
```diff
@@ -0,0 +1,274 @@
...
+			id: 'tagForAuthenticatedUser',
+			tasks: [
+				{ id: 'task1ForAuthenticatedUser' },
+				{ id: 'task2ForAuthenticatedUser' },
+				{ id: 'task3ForAuthenticatedUser' },
+			],
+		});
+
+		// If this is called more than once then we are not batching properly
+		expect(spyOnArtistDataProvider).toHaveBeenCalledTimes(1);
```
thekevinbrown, line 272: ❤️

thekevinbrown (approved):
Looks great, thanks! A couple small suggestions you can take or leave as you like.

PR #1688, opened 2025-03-17
`EXOGW-410` - GitHub 1574 add range filter option in the admin UI for date ranges and number ranges

juan871003:
https://exogee.atlassian.net/browse/EXOGW-410

When
image

Admin UI renders
image

Which allows users to filter by a range:
- lower than a number
- higher than a number
- between two numbers


thekevinbrown (changes requested):
Let's make the inputs in the dropdown match the styling of the other filter inputs in terms of size, corners etc, then this looks good to go, cheers!

thekevinbrown (approved):

PR #1689, opened 2025-03-17
`EXOGW-142` - Add .gitignore file generation and fix README typo

juan871003:
https://exogee.atlassian.net/browse/EXOGW-142

add .gitignore file during project initialisation.


src/packages/cli/src/init/template.ts
```diff
@@ -160,3 +160,266 @@ export const makeReadme = (projectName: string) => {
...
+
+# IntelliJ IDEA settings
+
+# Deployment files
+
+# Performance test results
+
+# Output folders
+lib/
+.graphweaver/
```
thekevinbrown, line 422: Let's keep this to just one .gitignore and have it only exclude the files we care about, e.g. we don't need to manage for Nuxt, DynamoDB, TernJS, FuseBox, Parcel etc. Just Node files, `lib`, and `.graphweaver`

thekevinbrown (changes requested):

thekevinbrown (approved):

PR #1690, opened 2025-03-17
`EXOGW-73` - Generated custom lambda names are too long and machine-specific

juan871003:
https://exogee.atlassian.net/browse/EXOGW-73

### DIFF
--- a/src/packages/cdk/src/examples/lambda.ts
+++ b/src/packages/cdk/src/examples/lambda.ts
@@ -68,7 +68,7 @@ export const graphweaverApp = new GraphweaverApp(rootStack, 'TestGraphweaverDock
 	},
 	lambda: {
 		packageName: '@exogee/graphweaver',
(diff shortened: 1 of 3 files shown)


thekevinbrown:
This doesn't seem to match the ticket's requirements, can you review?

src/packages/cdk/src/examples/lambda.ts
```diff
@@ -68,6 +68,7 @@ export const graphweaverApp = new GraphweaverApp(rootStack, 'TestGraphweaverDock
 	},
 	lambda: {
 		packageName: '@exogee/graphweaver',
+		functionName: 'Graphweaver API',
```
thekevinbrown, line 71: Given that the space isn't a valid character, we should probably do this for the example:

```suggestion
		functionName: 'Graphweaver_API',
```

thekevinbrown (approved):

PR #1699, opened 2025-03-19
Tests: "Fuzzer" prototype

mackcheesman:
This PR adds a new set of tests that run some very simplistic tests against the `rest-with-auth` example to test Graphweaver's security. These tests are static and don't provide much meaningful coverage. The intention is to improve this over time.
So far, these static tests:
- Attempt a really simple blind SQL injection
- Try querying for entities with a range of characters
- Check for information leaked about existing users through the example loginPassword mutation


.github/workflows/fuzzy-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+          sleep 5 &&
+          pnpm test-fuzz &&
+          killall node
+
+        env:
+          CI: true
+          DATABASE_HOST: 127.0.0.1
+          DATABASE_USERNAME: root
+          DATABASE_PASSWORD: root
+          AUTH_PUBLIC_KEY_PEM_BASE64: "LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUZrd0V3WUhLb1pJemowQ0FRWUlLb1pJemowREFRY0RRZ0FFcVFSUC9nem1ZdVJyR012UzJxeXpLaU05c0Z2aQpyWFRWVUsrMDBHaFFDa2NhdThOcWZsWG9nOEhyTkVsalkwWWpYcVVqOCs2ZDlySkEwTHo0NmFGSmp3PT0KLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0tCg=="
```
thekevinbrown, line 70: The public key isn't a secret really, but it'd be nice to put it in as a secret just because it's bound to the private key, and we'll be able to update both in one place if we need to change it.

thekevinbrown (changes requested):
Small change that'd be nice for maintainability. I agree with you on the types being able to be improved, and we can work on that after merge I reckon.

.github/workflows/fuzzy-tests.yml
mackcheesman: I've gone and made that change in the other two workflows as well, since this was a copy-paste 👍

mackcheesman: My mistake, I thought it was already in as a secret - either I've got the wrong name or it's not there. Could you check?

thekevinbrown (approved):
👍

PR #1749, opened 2025-04-02
`EXOGW418` - Added ability to type in date picker Admin UI component

mackcheesman:
The date picker component now lets you type in it, instead of having to click through the calendar pop-up. Dates can be entered as either a date in local format (according to the user's browser/system settings) or in YYYY-MM-DD format.

I've also added proper labels to the time input component, fixed the styling on some inputs to make them larger/stop them from shrinking, and fixed a bug that caused an error to appear if using a DateTime range filter and refreshing the page.


thekevinbrown (approved):
Thanks!

PR #1793, opened 2025-05-09
WIP: Exogw 431 create root backend id children client id children backend id fails because children with backend id get created twice

twestos:


src/packages/end-to-end/src/__tests__/api/auth/password/challenge.test.ts
twestos, line 1: Can you take a look at the changes I've added here.

I'm not certain if my changes are nerfing the quality of this test.

src/packages/end-to-end/src/__tests__/api/auth/password/challenge.test.ts
```diff
@@ -21,9 +21,25 @@ class TaskProvider extends BaseDataProvider<any> {
...
+class TagProvider extends BaseDataProvider<any> {
+	public async withTransaction<T>(callback: () => Promise<T>) {
+		return await callback();
+	}
+	async updateOne(pk: string, data: any) {
+		return data;
+	}
+	async findOne(data: any) {
+		console.log('FIND ONE', data);
+		throw new Error('Authentication Error: Expected Token.');
```
twestos, line 41: Can you take a look through these changes to the test I've made.

I'm not sure if this is circumventing the intent of the test.

I kept getting an error saying FindOne not implemented as we are using the BaseDataProvider class.

twestos: will remove the console.log()

thekevinbrown: This is all good. Basically previously we didn't call findOne here to complete this action, but now we do. Which is fine in this case.

PR #1801, opened 2025-05-16
EXOGW-445: Throw errors when graphql-codegen fails to load a GQL document

mackcheesman:
When you had a problem with your GraphQL, `graphweaver build` would appear to succeed anyway (and none of the generated graphql files would get built properly). This was related to [an issue with graphql-codegen/cli](https://github.com/dotansimha/graphql-code-generator/issues/9172) which has recently been fixed.

The only thing in this PR is an upgrade to the most recent release of the affected package and some small tweaks to the config file to make those errors actually get thrown. 

Note, the build will fail on the *first* error only. I tried tinkering with it to get it to find all the errors, but couldn't (immediately) figure out how to after a few minutes. I might come back to this later, but for now this will at least make it clear that something went wrong, which is an improvement.

### DIFF
--- a/src/packages/builder/package.json
+++ b/src/packages/builder/package.json
@@ -20,7 +20,7 @@
 		"@exogee/graphweaver-config": "workspace:*",
 		"@exogee/graphweaver-mikroorm": "workspace:*",
 		"@graphql-codegen/add": "5.0.3",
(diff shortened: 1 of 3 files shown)


twestos:
Looks good to me, top stuff Mack!

twestos (approved):

PR #1809, opened 2025-06-18
Feature / Beta support for SQL Server

thekevinbrown:
This PR adds:
- Support for SQL Server, both in introspection and normal use. SQL Server connections leverage the existing MikroORM adapter with the `tedious` driver.
- An example project that shows how to connect to SQL Server with Graphweaver.
- Introspection now provides a friendly name on the new datasource to make it clear to users how to do that.


twestos (approved):
🚀

PR #1824, opened 2025-07-01
Add env var and optional login parameter for extended token generation

eddiesholl:
Proof of concept for supporting a 'remember me on this device' type of login. For now this requires some testing


thekevinbrown:
Going to go ahead and assume this is abandoned for now. Would be happy to have this functionality if anyone wants to pick this back up and get it over the line.

PR #1875, opened 2025-08-07
Improve detail-panel form handling for relationships

eddiesholl:
A set of related changes to get a complex detail panel saving, in particular around relationship controls. The changes are targeting:
- filter the payload we send to the server between form submission and hitting the network with the mutation
- when dealing with relationships, make sure the values stored in the formic form state are in the right format. Many means an array, one means an object or null
- allow unlinking a many to one relationship by passing a null

I don't think any of these changes could be considered breaking in semver terms. I would consider them bug fixes. If someone was relying on the shape of form state returned by a combobox, then upgrading to this version could break something. There's also a noticeable difference in what we submit with a mutation, we now only supply fields that have changed. I don't think these justify a major version bump. Maybe a minor is the right way to go?

### DIFF
--- a/src/packages/admin-ui-components/src/detail-panel/fields/boolean-field.tsx
+++ b/src/packages/admin-ui-components/src/detail-panel/fields/boolean-field.tsx
@@ -1,7 +1,6 @@
 import { Field, FieldProps } from 'formik';
-import { SelectOption, ComboBox, SelectMode } from '../../combo-box';
+import { ComboBox, SelectMode, SelectOption } from '../../combo-box';
 import { EntityField } from '../../utils';
-import { useDataTransform } from '../use-data-transform';
 
 export const BooleanField = ({
 	field,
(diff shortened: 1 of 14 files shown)


src/packages/admin-ui-components/src/detail-panel/fields/boolean-field.tsx
```diff
@@ -12,16 +11,6 @@ export const BooleanField = ({
 	autoFocus: boolean;
 	disabled?: boolean;
 }) => {
-	// Before we go up to the server we need to change over to a boolean from an array of options
```
eddiesholl, line 15: This was just wrong and redundant

eddiesholl, line 15: This was just wrong and redundant

src/packages/core/src/metadata-service/resolver.ts
```diff
@@ -137,8 +137,12 @@ export const resolveAdminUiMetadata = (hooks?: Hooks) => {
 					fieldObject.relationshipType = RelationshipType.ONE_TO_MANY;
 
 					const relatedEntityField = Object.values(relatedObject.fields).find((field) => {
-						const fieldType = field.getType() as { name?: string };
-						return fieldType.name === (entity.target as { name?: string }).name;
+						const { isList: isOtherSideList, fieldType: fieldTypeOtherSide } =
```
eddiesholl, line 140: I couldn't get relationship fields to be detected at this point. This should be a more accurate matching check

thekevinbrown (approved):
Looks great, thank you very much!

PR #1881, opened 2025-08-11
Fix: Expose form state helpers to custom components

mackcheesman:
We're using Formik on the detail panel form. Right now, you can't properly connect to the rest of the form state with a custom component (Formik will explode on you if you try). This PR adds a bit to the `CustomFieldComponent` component by grabbing some helpers from Formik to pass directly to the component.


thekevinbrown (approved):
Thanks! Sorry to be a pain, I just know we're committing to this API is all.

PR #1888, opened 2025-08-11
Nullable mutation inputs

eddiesholl:
A small follow up change to address something I missed in the recent PR. This change decouples whether a field is required in an update mutation from its 'nullable' status, which has impacts on queries. We don't want to have to mark every field as nullable to let a thin mutation through, at the expense of having to null check every field on a query.

We now supply seperate keys to describe if a key is required for create or for update, this lets us be much more specific. And nullable is now only important values read by queries, and not used for mutations

### DIFF
--- a/src/packages/end-to-end/src/__tests__/ui/sqlite/components/combobox.test.ts
+++ b/src/packages/end-to-end/src/__tests__/ui/sqlite/components/combobox.test.ts
@@ -20,7 +20,7 @@ test('Check Select field shows correct number of selected items after adding add
 	await page.getByRole('cell', { name: '3', exact: true }).click();
 	await page
 		.locator('div')
-		.filter({ hasText: /^tracks\*3 Selected×$/ })
+		.filter({ hasText: /^tracks3 Selected×$/ })
 		.click({ delay: 1000 });
 	await page.getByRole('listbox').getByText('"40"').click();
 	await expect(page.locator('form')).toContainText('4 Selected');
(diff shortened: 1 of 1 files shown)


src/packages/admin-ui-components/src/detail-panel/fields/boolean-field.tsx
```diff
@@ -11,6 +12,15 @@ export const BooleanField = ({
 	autoFocus: boolean;
 	disabled?: boolean;
 }) => {
+	useDataTransform({
```
eddiesholl, line 15: I removed this code in the previous PR as I was trying to flatten out the values stored in the form. It turns out that change is a lot more invasive than I thought, the combobox is really hard wired to store arrays of items. I've tested this across a range of scenarios and this is actually needed

src/packages/auth/src/auth-utils.ts
```diff
@@ -301,6 +301,12 @@ export async function checkAuthorization<G = unknown>(
 
 		if (isEntityMetadata(relatedEntityMetadata)) {
 			// Now we have a related entity, let's check and make sure we have permission
+			
+			// If a to-many relationship is being reset to an empty array, we don't need to check permissions
```
eddiesholl, line 305: This is a fix I found I needed when submitting various payloads. In this case you remove the last item from a to many relationship, and an empty array is passed to the server, which causes `requiredPermissionsForAction` to throw

eddiesholl, line 305: This is a fix I found I needed when submitting various payloads. In this case you remove the last item from a to many relationship, and an empty array is passed to the server, which causes `requiredPermissionsForAction` to throw

PR #1916, opened 2025-08-21
Feature/relationship edits

eddiesholl:


src/examples/sql-server/package.json
```diff
@@ -23,7 +23,7 @@
 	},
 	"devDependencies": {
 		"@types/node": "22.13.14",
-		"graphweaver": "2.16.1",
+		"graphweaver": "workspace:*",
```
eddiesholl, line 26: This is the only reference to a fixed version of a package in the mono repo. I believe it leads to multiple copies of packages in node_modules, which breaks decorators and other metadata tricks

src/package.json
```diff
@@ -55,7 +55,8 @@
 			"@mikro-orm/knex": "6.4.16",
 			"@mikro-orm/postgresql": "6.4.16",
 			"@mikro-orm/mysql": "6.4.16",
-			"@mikro-orm/sqlite": "6.4.16"
+			"@mikro-orm/sqlite": "6.4.16",
+			"graphql": "16.11.0"
```
eddiesholl, line 59: This shouldn't have any real effect on graphweaver itself, as they already reference this version,  but it's necessary to make pulling in an app like Curby reliable.

src/packages/admin-ui-components/src/detail-panel/fields/relationship-field.tsx
```diff
@@ -68,10 +69,12 @@ export const RelationshipField = ({
 		}
 	);
 
+	const [inputValue, onInputChange] = useState('');
+
 	const options = (data?.result ?? []).map<SelectOption>((item): SelectOption => {
 		const label = relatedEntity.summaryField || relatedEntity.primaryKeyField;
 		return { label: item[label], value: item[relatedEntity.primaryKeyField] };
-	});
+	}).filter(item => inputValue?.toLowerCase().length > 0 ? item.label?.toLowerCase().includes(inputValue.toLowerCase()) : true);
```
eddiesholl, line 77: Implement really dumb client side filtering for relationships. A better solution might be to adjust the filter params passed to the server, like in the relationship filter. This solution just applies a client side match to filter results when you type in the combo box

src/packages/admin-ui-components/src/detail-panel/graphql.ts
```diff
@@ -7,7 +7,7 @@ export const generateUpdateEntityMutation = (
 ) => gql`
     mutation updateEntity ($input: ${entity.name}UpdateInput!){
       update${entity.name} (input: $input) {
-        ${generateGqlSelectForEntityFields(entity, entityByType)}
+        ${generateGqlSelectForEntityFields(entity.fields.filter((field) => !field.hideInDetailForm), entityByType)}
```
eddiesholl, line 10: There is a bug in generateGqlSelectForEntityFields right now when items are enabled for the list but not for edit or vice versa. This solution requires the caller to know exactly what they are supposed to filter on, based on if they are fetching for a list or edit scenario. It could also be implemented as extra boolean parameters passed to generateGqlSelectForEntityFields, or new properties on the entity class, like `entity.fieldsForLists` and `entity.fieldsForEditing`

src/packages/admin-ui-components/src/combo-box/component.tsx
```diff
@@ -92,7 +92,7 @@ export const ComboBox = ({
 	});
 
 	// Clear typed text on blur if no item was selected
-	const handleBlur = () => {
+	const handleBlur = useCallback(() => {
```
eddiesholl, line 95: This seems to have been a performance issue, seems like a nice improvement

PR #1967, opened 2025-09-01
Fix / Show selected items in combo box

thekevinbrown:
Our UX around selected items is not great at the moment. If one item is selected, it's easy to tell which it is, but once multiple are selected then it is really hard as a user to see what you've chosen.

This PR provides a better UX for this by showing the selected items at the top of the dropdown, always scrolling to the top of the dropdown when it opens, and allowing users to individually deselect options instead of the all or nothing clear we have currently:

**With 3 selected items**
Screenshot 2025-09-01 at 12 08 44 pm

**When hovering a selected item**
Screenshot 2025-09-01 at 12 08 59 pm

This applies to the `ComboBox`, `RelationshipFilter`, and `RelationshipField` components.

### DIFF
--- a/.github/workflows/postgres-end-to-end-tests.yml
+++ b/.github/workflows/postgres-end-to-end-tests.yml
@@ -81,7 +81,7 @@ jobs:
         if: ${{ !cancelled() }}
         with:
           name: playwright-report
-          path: playwright-report/
+          path: src/packages/end-to-end/playwright-report/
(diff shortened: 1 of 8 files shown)


thekevinbrown:
Going to go ahead and merge this with known SonarQube flags around accessibility, as the current component has these same problems. We'll work on fixing those next.

PR #1998, opened 2025-09-04
Fix: ComboBox improvements

mackcheesman:
## Changes to the `ComboBox` component:
- Used the `useMultipleSelection` hook from downshift to restructure how the selected options were being handled and rendered
- The input part of the component now always renders, regardless of if typing is allowed or if something has been selected.
  - This lets the browser handle default behaviour with the keyboard (in particular, using the arrow keys to go up and down the list).
- Added focus styling to the 'clear all' button on the option pill (it's not perfect, but it at least makes a bit clearer when you're tabbing through the form:
Screenshot 2025-09-04 at 11 20 59 am

- Got rid of the CSS rule on pointer events on the toggle button. AFAICT it was added to prevent a double click that sometimes happens, but the button wasn't actually doing anything with that rule.

## Other changes
- Added one more test to cover a bit with multiple selection mode.
- There's other changes I was in the midst of trying to make that I've rolled back for now. They'll come in a follow-up PR to this.


thekevinbrown (approved):
This is awesome, thank you!

PR #2061, opened 2025-09-12
Chore: Disabling deprecation warning over module resolution (for now)

mackcheesman:
Having `"moduleResolution": "node"` in tsconfig.json now results in a deprecation warning that looks like an error.

I tried switching the module resolution option to bundler, but that didn't immediately work. It's something that we could possibly untangle, but for now the error is just acting as unhelpful noise, so I'm silencing it.


mackcheesman:
This is clearly not the 5 minute fix I thought it was going to be - I'm going to come back to this later when I have time to dig through things properly.

PR #2597, opened 2026-09-14
Feature / Support TypeScript config files

thekevinbrown:
- Added support for TypeScript config files.
- Changed the examples to use .ts instead of .js configs.


thekevinbrown:
This duplication doesn't concern me given what we're doing.

